PR #6, opened 2021-01-13
Adding Psi4

sheepforce:
This PR attempts to add Psi4 and its dependencies. Most expressions are taken from my [Chemix](https://gitlab.com/theoretical-chemistry-jena/nixwithchemistry) overlay. 

Packages to add:
- [x] Psi4
- [x] QCEngine
- [x] QCElemental
- [x] Gau2Grid
- [x] DKH
- [x] DFT-D3
- [x] LibEFP and PyLibEFP
- [x] PCMSolver

See also the discussion in the [NixOS Discourse](https://discourse.nixos.org/t/mixed-python-c-with-python-runtime-dependencies/7679). As mentioned in the forum, many packages required pinning and tweaks on the CMake flags to get Psi4 running.

If we find nicer versions to get Psi4's Cmake into finding the external dependencies, I would be happy :slightly_smiling_face:


markuskowa:
As a starting point here, since this an overlay: it may be easier to test if we agree on a version of nixpkgs to test against. My internal Hydra successfully ran the overlay's master branch against nixpkgs-unstable revision f18ba04.

sheepforce:
Alright :slightly_smiling_face: I don't have a Hydra server at hand, but I will build against this commit on my locale machine then. At the moment I was using my nixos-20.09 channel.

sheepforce:
OK, I am done with adding the packages. They all build properly with nixpkgs-unstable f18ba04. A few things you might want to have a look at (in no specific order)

- Your Chemps2 now works with other BLAS implementations
- I changed `overrideDerivation` to `overrideAttributes` for `libint`. The former version did not allow multiple overrides. Furthermore I made the angular momentum an optional argument to the function.
- I did dirty things to CMake in a few places. These include the `libefp` derivation itself and many places in Psi4. I basically override every default CMake argument, that came from Nixpkgs. I can't get it to work otherwise. Unfortunately I also have no idea what causes the problem, when using normal Nixpkgs CMake arguments.
- Psi4 has a huge test suite and I could never convince it to pass completely (also not outside of Nix). Most tests pass. Others are extremely wrong. However, significantly more tests seem to fail when using another BLAS/LAPACK provider than MKL. I would almost consider allowing only MKL as BLAS/LAPACK provider here. It also builds and works with OpenBLAS and I haven't encountered unphysical behaviour when actually using it, but the test suite concerns me. Performance is also significantly better with MKL.

default.nix
```diff
@@ -223,6 +223,15 @@ in with super;
...
   }).overrideAttrs (x: { preCheck = "export OMP_NUM_THREADS=4"; });
 
+  dkh = callPackage ./dkh {};
+
+  dftd3 = callPackage ./dft-d3 {};
+
+  libefp = callPackage ./libefp {};
+
+  pcmsolver = callPackage ./pcmsolver {};
+  pcmsolver-1_2_3 = callPackage ./pcmsolver/1_2_3.nix {};
```
markuskowa, line 233: Where is pcmsolver-1.2.3 used/where is it needed?

dft-d3/default.nix
```diff
@@ -0,0 +1,39 @@
...
+    '';
+
+    hardeningDisable = [
+      "format"
+    ];
+
+    meta = with stdenv.lib; {
+      description = "Dispersion correction for DFT";
+      homepage = "https://www.chemie.uni-bonn.de/pctc/mulliken-center/software/dft-d3/get-the-current-version-of-dft-d3";
+      platforms = platforms.unix;
```
markuskowa, line 37: Just for completeness: can you add a license here?

dft-d3/default.nix
```diff
@@ -0,0 +1,39 @@
...
+    version = "3.2rev0";
+
+    nativeBuildInputs = [ gfortran ];
+
+    src = fetchurl {
+      url = "https://www.chemie.uni-bonn.de/pctc/mulliken-center/software/dft-d3/dftd3.tgz";
+      sha256 = "0n8gi0raz8rik1rkd4ifq43wxs56ykxlhm22hpyq3ak1ixszjz6r";
+    };
+
+    unpackPhase = ''
```
markuskowa, line 13: Is this really needed? A `tgz` should be handled by the default `unpackPhase`?

pythonPackages.nix
```diff
@@ -7,4 +7,12 @@ in {
   pyscf = callPackage ./pyscf { };
   pyquante = callPackage ./pyquante { };
   pychemps2 = callPackage ./chemps2/PyChemMPS2.nix { };
+  qcelemental = callPackage ./qcelemental { };
+  qcengine = callPackage ./qcengine { };
+  gau2grid-1_3_1 = callPackage ./gau2grid/1_3_1.nix { };
```
markuskowa, line 12: As a general comment for the cases, where we have multiple versions of a package: as far as I can see, the derivations are identical except for `version` and `sha256`. I would be better to just have one derivation file instead and make `version` and `sha256` input parameters, with defaults for the corresponding parameters for the latest stable (I know should do the same for all my molpro derivations).

pythonPackages.nix
```diff
@@ -7,4 +7,12 @@ in {
...
+  gau2grid-2_0_4 = callPackage ./gau2grid/2_0_4.nix { };
+  gau2grid = callPackage ./gau2grid { };
+  pylibefp = callPackage ./pylibefp { };
+  psi4 = callPackage ./psi4 { };
+  psi4-9b60184 = callPackage ./psi4/9b60184.nix { };
```
markuskowa, line 17: I assume this an unstable/master branch version of psi4? I would change the naming scheme to be in line with what nixpkgs recommends:
* name the attribute `psi4Unstable`.
* name the nix file `unstable.nix`
* `version` should be the date of the commit.

psi4/9b60184.nix
```diff
@@ -0,0 +1,244 @@
...
+, pcmsolver
+, libefp
+, chemps2
+, hdf5
+, hdf5-cpp
+# Optional
+, mrcc ? null
+}:
+let
+  useMKL = blas.passthru.implementation == "mkl";
```
markuskowa, line 30: openblas vs. MKL: bagel also has the problem that some routines produce error when compiled against openblas (i.e. it only works properly with MKL). 
Here's what would suggest:
* Keep using the generic `lapack` and `blas` inputs but add an assert that fails in case it is not MKL.
* Call the package with something like this:
```
psi4 = callPackage { blas = blas.override { blasProvider = mkl; }; lapack = .... };
```

psi4/9b60184.nix
```diff
@@ -0,0 +1,244 @@
...
+  blas_ = blas.passthru.provider;
+  lapack_ = lapack.passthru.provider;
+
+  # Psi4 requires some special cmake flags. Using Nix's defaults usually does not work.
+  specialInstallCmakeFlags = [
```
markuskowa, line 35: I will have a look at it and see if can come up with a solution.

psi4/9b60184.nix
```diff
@@ -0,0 +1,244 @@
...
+      libefp
+      deepdiff
+    ] ++ qcelemental.passthru.requiredPythonModules
+    );
+
+    postFixup = ''
+      # Make libraries and external binaries available
+      wrapProgram $out/bin/psi4 \
+        --prefix PATH : ${binSearchPath} \
+        --prefix PYTHONPATH : ${libSearchPath} \
```
markuskowa, line 217: This could be done shorter by using `python.withPackages`. Here's the basic idea:
```
 buildInputs = [ python' ];
 python' = python.withPackages(ps: with ps; [ numpy  pybind11 qcelemental libefp deepdiff])
```
That way you get a wrapped python interpreter that has the packages already in its `PYTHONPATH`.

markuskowa (commented):
Here's my first round of comments. I think in general it looks pretty good. I have not tried to build it yet locally.

default.nix
sheepforce: It is not as far as i can see. I have removed it. Just copied it from my overlay.

dft-d3/default.nix
sheepforce: Done. Actually none is given by Grimme, so I guess the best estimate is `unfree`?

dft-d3/default.nix
sheepforce: Unfortunately not. `unpackPhase produced no directory` or something like this happens.

pythonPackages.nix
sheepforce: Has been done.

pythonPackages.nix
sheepforce: It is a snapshot, from master, yes. Have adapted it.

psi4/9b60184.nix
sheepforce: I have added the assert for MKL. But as psi4 lives in the `pythonPackages.nix` and its `super` does not know BLAS, I cannot make this override there, as far as I can see. 

 The unwrap of the BLAS provider, that I am doing there with `blas_ = blas_ = blas.passthru.provider` and equivalently for LAPACK, is unfortunately required. Psi4's CMake misses some information about OpenMP libraries, when directly using `blas` and fails to link.

I view of both problems, it might be reasonable to directly pass MKL to Psi4? Alternatively i could use some hint how to do your suggested override in `pythonPackages.nix`.

psi4/9b60184.nix
sheepforce: I would prefer not to do this :slightly_smiling_face:. It actually caused some problems in our applications. Psi4 does not simply require python modules at runtime, but is actually a python module, that can be imported with `import psi4` in python. A colleague writes a python program - [pysisyphus](https://github.com/eljost/pysisyphus) - implementing advanced optimisers. It wraps several QC packages and for Psi4 it uses the python library. Now when we build the derivation of pysisyphus with nix and Psi4 is one of the `propagatedBuildInputs` to pysis and one uses the `python.withPackages` stuff, the derivations fail to build, as `python.withPackages` seems to create an extra derivation with symlinkJoin or something? In the end there are multiple versions of shared dependencies like numpy and python gets confused.

Long story short: i think the `python.withPackages` version causes some problems, when using Psi4 as a python module, rather than as an executable.

dft-d3/default.nix
markuskowa: Yes, it should be `unfree` nothing else specifed.

psi4/9b60184.nix
markuskowa: Oh I did not realise that is defined as python package. In that case: did you ever try to build it with `buildPythonPackage` instead of `mkDerivation`? That should take of most the python environment issues.

sheepforce:
> I find the test suites of many QC program hard to use. It is not easy to judge what it means if a test actually fails. Is it an insignificant deviation from the assert value or is it serious problem. The way most of them are written, they are only useful for the developers. I try to use minimal but representative inputs for testing and check for final energies. This makes it easier to see if something broke after an update.

You are probably right :smile: In this case I would suggest a small DF-OO-MP2.5 optimisation. The dfocc module seemed especially problematic with respect to threading of BLAS. 
```
memory 1 GB
molecule Water {
 0 1
 O
 H 1 0.8
 H 1 0.8 2 110
}

set {
 basis cc-pvdz
 scf_type df
 mp_type df
 reference rhf
}

optimize("omp2.5")
```


> On which CPUs are you running it?

It is an Intel Xeon W-2155 workstation CPU. But the problem with OpenBLAS in Psi4 is not so much the CPU architecture optimisation but the threading behaviour. You only get correct behaviour in dfcocc and some other modules with MKL (OpenMP parallel) or with OpenBLAS with `OMP_NUM_THREADS=1`, which makes it simply slow by not being parallel.

psi4/9b60184.nix
markuskowa: Same as above:
Oh I did not realise that is defined as python package. In that case: did you ever try to build it with buildPythonPackage instead of mkDerivation? That should take of most the python environment issues.

markuskowa:
> You are probably right ￼ In this case I would suggest a small DF-OO-MP2.5 optimisation. The dfocc module seemed especially problematic with respect to threading of BLAS.

The test example sounds reasonable. We could add it as separate test to have confirmation that is still works after an update.

> It is an Intel Xeon W-2155 workstation CPU. But the problem with OpenBLAS in Psi4 is not so much the CPU architecture optimisation but the threading behaviour. You only get correct behaviour in dfcocc and some other modules with MKL (OpenMP parallel) or with OpenBLAS with OMP_NUM_THREADS=1, which makes it simply slow by not being parallel.

This has been a problem in other contexts too. There are several issues for that. See for example: https://github.com/NixOS/nixpkgs/issues/59901. IIRC on some occasion that problem was coupled to specific SSE/AVX kernels.

psi4/9b60184.nix
markuskowa: To solve override we would need to fix the construction of how the override set is pulled in, which is probably a good idea. I ran into the some problem in another context.

This may get the job done(?):

`pythonPackages.nix`:
```
self: super: selfPkgs: superPkgs:
```
`default.nix`:
```
pythonOverrides = (import ./pythonPackages.nix) self super;
```

sheepforce: > Oh I did not realise that is defined as python package. In that case: did you ever try to build it with `buildPythonPackage` instead of `mkDerivation`? That should take of most the python environment issues.

I gave it a quick try but was not too successfull. It actually is more a C++/Cmake project with python glue around it. There is no `setup.py` or anything similar, so i guess there would be lots of fiddling around with buildPythonPackage?

> This may get the job done(?):
Yes, this is a good idea and its working :slightly_smiling_face:

sheepforce:
The test case is added and as far as I can tell everything seems to work so far.

rmcgibbo:
@sheepforce: I just tried building this from 0d4744f, and got
```
hash mismatch in fixed-output derivation '/nix/store/ykj5x09a65f6ymyiapmv1zxv0jdlbvmg-source':
  wanted: sha256:0jcvl3chni4f0hddx9blaia3kccfqx7cszrwavp0a35d42n0x5i2
  got:    sha256:1zlihc7hp2vcdcwi82n7xg41bm1vv7n1mx1jdff4w1amhg9rlahm
cannot build derivation '/nix/store/y7dvy3rrj22nvb8qnrsm1nj4b9brxmb6-libefp-1.5.0.drv': 1 dependencies couldn't be built
building '/nix/store/k03vin4b9im2w6p5harm3knbmq1v0zjc-source.drv'...
building '/nix/store/pfsy8cskqsk3c2rrw9gjz3i3jshgzibm-source.drv'...
building '/nix/store/plvrzbb8006gx33zsydknyr9klf8css9-source.drv'...
building '/nix/store/rxxlvcpw6wiaaxc70m5cq21znz3kx0c4-source.drv'...
building '/nix/store/cddgyr4c5hhmhyzwdpnlxkigf4ssg217-splines.tar.xz.drv'...
building '/nix/store/drz9fixsav9mw3x4dz2s0k2hcaa2i9hl-stringenc.tar.xz.drv'...
building '/nix/store/7i7fyij7laazsjl0cyf54syp77b0p2xb-subfig.tar.xz.drv'...
building '/nix/store/88vz6y4cnbxkx4ryb80cyhhrxmf5jlnc-tetragonos.tar.xz.drv'...
cannot build derivation '/nix/store/f1vidvz3xjj4ipp88gfvgnb8j2zcqnan-psi4-1.3.2.drv': 1 dependencies couldn't be built
cannot build derivation '/nix/store/3a02qp05yhvgqfs6bbp126b7nqdrxn7k-home-manager-path.drv': 1 dependencies couldn't be built
cannot build derivation '/nix/store/h5r6cn56ixhzv7mfwfsghz9swqk63n0x-home-manager-generation.drv': 1 dependencies couldn't be built
error: build of '/nix/store/h5r6cn56ixhzv7mfwfsghz9swqk63n0x-home-manager-generation.drv' failed
```
```

**Comment by rmcgibbo** [comment]
Also, in addition to providing `psi4` inside pythonPackages accessible as `python3.pkgs.psi4`, you might consider also adding a buildPythonApplication in the root `default.nix`, so that users who don't need psi4 inside of a custom python environment with other packages could just add `environment.systemPackages = with pkgs; [psi4]` to their configs.

File: psi4/default.nix [hunk 9]
```diff
@@ -238,6 +262,12 @@ in
       ctest -j $NIX_BUILD_CORES
     '';
 
+    doInstallCheck = true;
+    installCheckPhase = ''
+      export OMP_NUM_THREADS=$NIX_BUILD_CORES
```
markuskowa, line 267: I would fix the number of core for the test to two. That makes the test more reproducible.
```
export OMP_NUM_THREADS=2
```

sheepforce: Good idea, done.

sheepforce:
> @sheepforce: I just tried building this from [0d4744f](https://github.com/markuskowa/NixOS-QChem/commit/0d4744f00d9cf6d9f97f2808b65649d265bdb37d), and got
> 
> ```
> hash mismatch in fixed-output derivation '/nix/store/ykj5x09a65f6ymyiapmv1zxv0jdlbvmg-source':
>   wanted: sha256:0jcvl3chni4f0hddx9blaia3kccfqx7cszrwavp0a35d42n0x5i2
>   got:    sha256:1zlihc7hp2vcdcwi82n7xg41bm1vv7n1mx1jdff4w1amhg9rlahm
> cannot build derivation '/nix/store/y7dvy3rrj22nvb8qnrsm1nj4b9brxmb6-libefp-1.5.0.drv': 1 dependencies couldn't be built
> building '/nix/store/k03vin4b9im2w6p5harm3knbmq1v0zjc-source.drv'...
> building '/nix/store/pfsy8cskqsk3c2rrw9gjz3i3jshgzibm-source.drv'...
> building '/nix/store/plvrzbb8006gx33zsydknyr9klf8css9-source.drv'...
> building '/nix/store/rxxlvcpw6wiaaxc70m5cq21znz3kx0c4-source.drv'...
> building '/nix/store/cddgyr4c5hhmhyzwdpnlxkigf4ssg217-splines.tar.xz.drv'...
> building '/nix/store/drz9fixsav9mw3x4dz2s0k2hcaa2i9hl-stringenc.tar.xz.drv'...
> building '/nix/store/7i7fyij7laazsjl0cyf54syp77b0p2xb-subfig.tar.xz.drv'...
> building '/nix/store/88vz6y4cnbxkx4ryb80cyhhrxmf5jlnc-tetragonos.tar.xz.drv'...
> cannot build derivation '/nix/store/f1vidvz3xjj4ipp88gfvgnb8j2zcqnan-psi4-1.3.2.drv': 1 dependencies couldn't be built
> cannot build derivation '/nix/store/3a02qp05yhvgqfs6bbp126b7nqdrxn7k-home-manager-path.drv': 1 dependencies couldn't be built
> cannot build derivation '/nix/store/h5r6cn56ixhzv7mfwfsghz9swqk63n0x-home-manager-generation.drv': 1 dependencies couldn't be built
> error: build of '/nix/store/h5r6cn56ixhzv7mfwfsghz9swqk63n0x-home-manager-generation.drv' failed
> ```
@rmcgibbo, oh you are right. I somehow messed up git commits and hashes. Could you try again with [
8c2dca5 ](https://github.com/markuskowa/NixOS-QChem/pull/6/commits/8c2dca55ed2bcb8cc13546789ee5778ce407b4c3)? :slightly_smiling_face: 

> Also, in addition to providing psi4 inside pythonPackages accessible as python3.pkgs.psi4, you might consider also adding a buildPythonApplication in the root default.nix, so that users who don't need psi4 inside of a custom python environment with other packages could just add environment.systemPackages = with pkgs; [psi4] to their configs.


I have to admit I don't get what advantage you would gain from this being on the top level in addition. What is wrong with
```
environment.systemPackages = with pkgs; [
  python3Packages.psi4
];
```
? :slightly_smiling_face: 
As mentioned above I have a hard time with getting Psi4 to run with `buildPythonApplication` as this project is more driven by CMake rather than Python's setuptools. However, I do not build a custom set of Python packages, as exactly this was causing problems in my usecase. Therefore it should perfectly behave as if built with `buildPythonPackage`.

markuskowa:
> I have to admit I don't get what advantage you would gain from this being on the top level in addition. What is wrong with

If it is an application and not a library/pythonPackage then it should be a top-level package in `pkgs`. If it is under `pythonXPackages the user needs to decide which python version is the right one. `pythonXPackages` is a place for python libraries. 
I also would expect that `environment.systemPackages = with pkgs; [  python3Packages.psi4 ];` is potential source of trouble: now you have a python interpreter (and python include paths?) in your global environment, which may collide with with other apps and packages (and other python versions).

I will give it try later today and see if I can come up with a fix or idea for the `cmake` issue.

rmcgibbo:
Perhaps `toPythonApplication` can be used to wrap the library as an application. It may be as simple as
```
psi4 = toPythonApplication python3Packages.psi4;
```

rmcgibbo:
@sheepforce, I'm still having some mismatches with the sha256s:
```
hash mismatch in fixed-output derivation '/nix/store/0s3ncb296vpdg8gni1c3wm3rff74f8bl-source':
  wanted: sha256:0zkfil7cxjip79wqvhljk1ifjq0cwxzx6wlxgp63b6wbagma0i12
  got:    sha256:1c01flf0xlm68dsv0a4n6bxglyj3x5mj8shmxglfpywbf74i3vnb
cannot build derivation '/nix/store/4wzsizf5pdsx8795i0x19c0gvig85rps-gau2grid-2.0.7.drv': 1 dependencies couldn't be built
```
with
```
  (import (builtins.fetchTarball {
    url = "https://github.com/sheepforce/NixOS-QChem/archive/8c2dca55ed2bcb8cc13546789ee5778ce407b4c3.tar.gz";
    sha256 = "130l47jmgs0564767498ak2wr28afszzpbsqracdzgmdf4zp8ll2";
  }))
  ```
  
  
FWIW I was adding it to my home.nix (home-manager) with
```
home.packages = [
    ...
  (python3Packages.toPythonApplication python3Packages.psi4)
]
```
so I think that strategy will work fine.

File: libefp/default.nix [hunk 10]
```diff
@@ -0,0 +1,76 @@
...
+    ];
+
+    src =
+      let
+          rev = if psi4_patches
+                  then "15cd7ce91239c04b5c32ed101bde6cc36c57550a"
+                  else version;
+          sha256 = if psi4_patches
+                     then "1zlihc7hp2vcdcwi82n7xg41bm1vv7n1mx1jdff4w1amhg9rlahm"
+                     else "0jcvl3chni4f0hddx9blaia3kccfqx7cszrwavp0a35d42n0x5i2";
```
markuskowa, line 38: The hashes are maybe swaped? I get:
```
unpacking source archive /build/1.5.0.tar.gz
hash mismatch in fixed-output derivation '/nix/store/ykj5x09a65f6ymyiapmv1zxv0jdlbvmg-source':
  wanted: sha256:0jcvl3chni4f0hddx9blaia3kccfqx7cszrwavp0a35d42n0x5i2
  got:    sha256:1zlihc7hp2vcdcwi82n7xg41bm1vv7n1mx1jdff4w1amhg9rlahm
```

pcmsolver/default.nix
```diff
@@ -0,0 +1,45 @@
...
+      boost
+      eigen
+      zlib
+    ];
+
+    src = fetchFromGitHub  {
+      owner = "PCMSolver";
+      repo = pname;
+      rev = "v${version}";
+      sha256= "117f9ic41xagyzkfwa1sbfcmycc8nlm6ivflkdc228xmp7axvf9m";
```
markuskowa, line 28:
```
unpacking source archive /build/v1.3.0.tar.gz
hash mismatch in fixed-output derivation '/nix/store/16z3y6fdr4sdbrg3m7h5y3hqmiahygxq-source':
  wanted: sha256:117f9ic41xagyzkfwa1sbfcmycc8nlm6ivflkdc228xmp7axvf9m
  got:    sha256:0jrxr8z21hjy7ik999hna9rdqy221kbkl3qkb06xw7g80rc9x9yr
```

pythonPackages.nix
```diff
@@ -7,4 +7,18 @@ in {
   pyscf = callPackage ./pyscf { };
   pyquante = callPackage ./pyquante { };
   pychemps2 = callPackage ./chemps2/PyChemMPS2.nix { };
+  qcelemental = callPackage ./qcelemental { };
+  qcengine = callPackage ./qcengine { };
+  gau2grid-1_3_1 = callPackage ./gau2grid { version = "1.3.1"; sha256 = "0zkfil7cxjip79wqvhljk1ifjq0cwxzx6wlxgp63b6wbagma0i12"; };
+  gau2grid-2_0_4 = callPackage ./gau2grid { version = "2.0.4"; sha256 = "0qypq8iax0n6yfi4223zya468v24b60nr0x43ypmsafj0104zqa6"; };
```
markuskowa, line 13:
```
unpacking source archive /build/v2.0.7.tar.gz
hash mismatch in fixed-output derivation '/nix/store/0s3ncb296vpdg8gni1c3wm3rff74f8bl-source':
  wanted: sha256:0qypq8iax0n6yfi4223zya468v24b60nr0x43ypmsafj0104zqa6
  got:    sha256:1c01flf0xlm68dsv0a4n6bxglyj3x5mj8shmxglfpywbf74i3vnb
```

pythonPackages.nix
```diff
@@ -7,4 +7,18 @@ in {
   pyscf = callPackage ./pyscf { };
   pyquante = callPackage ./pyquante { };
   pychemps2 = callPackage ./chemps2/PyChemMPS2.nix { };
+  qcelemental = callPackage ./qcelemental { };
+  qcengine = callPackage ./qcengine { };
+  gau2grid-1_3_1 = callPackage ./gau2grid { version = "1.3.1"; sha256 = "0zkfil7cxjip79wqvhljk1ifjq0cwxzx6wlxgp63b6wbagma0i12"; };
```
markuskowa, line 12:
```
unpacking source archive /build/v2.0.7.tar.gz
hash mismatch in fixed-output derivation '/nix/store/0s3ncb296vpdg8gni1c3wm3rff74f8bl-source':
  wanted: sha256:0zkfil7cxjip79wqvhljk1ifjq0cwxzx6wlxgp63b6wbagma0i12
  got:    sha256:1c01flf0xlm68dsv0a4n6bxglyj3x5mj8shmxglfpywbf74i3vnb
```

libefp/default.nix
sheepforce: You are right. They were. I must have messed up completely with all the hashes when trying to merge everything into one file. Confirmed with `nix-prefetch-url` and should be fixed now by [fc855a8](https://github.com/markuskowa/NixOS-QChem/pull/6/commits/fc855a8168bcab13bcb2ab74deb65b844047c6c8).

pcmsolver/default.nix
sheepforce: same as with libefp. I am sorry. It didn't show up anymore as soon as I had it in the nix store and the hash existed somewhere. Checked with `nix-prefetch-url` and should be fixed by [37a969f](https://github.com/markuskowa/NixOS-QChem/pull/6/commits/37a969fa56e77e3e69d5eb37b835383c2644f282).

pythonPackages.nix
sheepforce: Always build version 2.0.7 due to stupid error (overriding the version argument from the function in the derivation) and hence the mismatches for other versions. Fixed by [https://github.com/markuskowa/NixOS-QChem/pull/6/commits/67d2f28b597671907ff57e51cfb280605c2bef35](https://github.com/markuskowa/NixOS-QChem/pull/6/commits/67d2f28b597671907ff57e51cfb280605c2bef35)

pythonPackages.nix
sheepforce: same as above, should be fixed by [67d2f28](https://github.com/markuskowa/NixOS-QChem/pull/6/commits/67d2f28b597671907ff57e51cfb280605c2bef35)

sheepforce:
@rmcgibbo 
> Perhaps `toPythonApplication` can be used to wrap the library as an application. It may be as simple as
> 
> ```
> psi4 = toPythonApplication python3Packages.psi4;
> ```
> 
> To expose both the Psi4 python library as a library and the application wrapper as an application.

Wow, why didn't i come up with this? :sweat_smile: It required Psi4 to be build with `buildPythonPackage` as it otherwise would complain about some missing name attribute. I had a look now again at `buildPythonPackage` and actually it was not too difficult to build with it instead of `mkDerivation`. Therefore we now have Psi4 in `python3Packages.psi4` and as a python application on top level. :slightly_smiling_face: 
([9ccfbeb](https://github.com/markuskowa/NixOS-QChem/pull/6/commits/9ccfbebb33ea28847fdf2dcb00e6d025f5efe4ce))

--------------

@markuskowa , sorry for the mess with the hashes. I hope that i finally fixed them now. I double checked with `nix-prefetch-url`. I didn't notice the problem as i had all the files in the store and nix simply used them. Therefore I never got any error.

sheepforce:
A small problem that I couldn't figure out right now for Psi4, is that the `installCheckPhase` is not executed anymore by `buildPythonPackage`. So the test does not run at the moment.

markuskowa:
@sheepforce thanks! I can build it now. I will check the cmake issues and see what we can do with the test.

psi4/unstable.nix
```diff
@@ -0,0 +1,292 @@
+{ lib, buildPythonPackage, buildPackages, makeWrapper, fetchFromGitHub, pkg-config, writeTextFile
```
markuskowa, line 1: The derivation for unstable is identical to the stable version, expect for one dependency. I suggest to put in one file for now.

markuskowa (commented):
Just some minor suggestions left here. Can you please also squash the commits so that we have ~1 commit for each package addtion (or package fix). I think we can merge it after that.

 I looked at the `cmake` issue. This is a bit of mess in psi4 itself so I can not come up with better solution for now. Let's keep it like it is since you have figured out how to make it work. It's not pretty in a nix sense but it seems to get job done. 
Regarding the test. It is ok to leave it where it is now. I will take it out later create a separate test for it.

libefp/default.nix
```diff
@@ -0,0 +1,76 @@
...
+    ];
+
+    buildInputs = [
+      blas
+      lapack
+    ];
+
+    src =
+      let
+          rev = if psi4_patches
```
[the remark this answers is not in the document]
sheepforce: Done.

psi4/unstable.nix
sheepforce: Done

sheepforce:
Alright :slightly_smiling_face: I have implemented your requests. I will squash the commits and then force push.

markuskowa:
Thank you! that looks good now. I'll wait for your squash.

sheepforce:
Its done :slightly_smiling_face:

markuskowa:
Completed. Thanks for the contribution!

sheepforce:
Thank your two for your valuable hints. I will take some inspiration from them. :slightly_smiling_face:

PR #9, opened 2021-04-23
Turbomole 7.5

sheepforce:
Addition of Turbomole 7.5. Turbomole requires heavy bending of it's directory structure to work with Nix. 
The original distribution requires to execute some scripts, which set environment variables and decide which executables are to be used. This Nix derivations wraps the executables with proper environments to avoid this error prone procedure and the Turbomole flavour (MPI vs SMP) can be selected by an argument to the derivation.

A test case has been added, but I guess it should not always be executed? Maybe cfg.nix needs an additional attribute for the availability of Turbomole and other packages. 

Regarding the sha256 of the source: we have the Turbomole developers here in Jena and have versions directly from them. It might be that the hash of the version you can get from their website is different, but the release is exactly 7.5.


tests/turbomole/default.nix
```diff
@@ -0,0 +1,33 @@
...
+  testScript = ''
+    @test "RI-ADC2" {
+      chmod +w auxbasis
+      chmod +w basis
+      chmod +w control
+      chmod +w coord
+      chmod +w mos
+      ${turbomole}/bin/ridft -smpcpus 2 > ridft.out
+      ${turbomole}/bin/ricc2 -smpcpus 2 > ricc2.out
+      grep "HOMO-LUMO gap:    0.48876" ridft.out
```
markuskowa, line 30: Maybe a simple check also for ricc2?

turbomole/default.nix
```diff
@@ -0,0 +1,147 @@
...
+, gcc-unwrapped
+, which
+, less
+, more
+, coreutils
+, useMPI ? false
+}:
+assert
+  lib.asserts.assertMsg
+  (stdenv.system == "x86_64-linux")
```
markuskowa, line 15: I don't think  we need this. Just set `meta.platform`.

turbomole/default.nix
```diff
@@ -0,0 +1,147 @@
+{ stdenv, lib, makeWrapper, requireFile, autoPatchelfHook, writeTextFile
+, perl
```
markuskowa, line 2: Uniform formatting: use either one-per-line or multiple per line.

turbomole/default.nix
```diff
@@ -0,0 +1,147 @@
...
+let
+  systemName = if stdenv.isLinux && stdenv.isx86_64 then "em64t-unknown-linux-gnu" else "x86_64-unknown-linux-gnu";
+
+  # Meta information
+  pname = "turbomole";
+  version = "7.5";
+
+
+in stdenv.mkDerivation rec {
+  pname = "turbomole";
```
markuskowa, line 27: Use `inherit` for `pname` and `version`.

turbomole/default.nix
```diff
@@ -0,0 +1,147 @@
...
+        --set TURBOARCH ${systemName}
+    done;
+  '';
+
+  meta = with lib; {
+    description = "General purpose quantum chemistry program. Tools, not Toys!";
+    homepage = "https://www.turbomole.org/";
+    license = licenses.unfree;
+    platforms = platforms.linux;
+    architectures = [ "amd64" ];
```
markuskowa, line 145: This can be removed.

turbomole/default.nix
```diff
@@ -0,0 +1,147 @@
...
+        --set PARA_ARCH "${PARA_ARCH}" \
+        --set TURBOARCH ${systemName}
+    done;
+  '';
+
+  meta = with lib; {
+    description = "General purpose quantum chemistry program. Tools, not Toys!";
+    homepage = "https://www.turbomole.org/";
+    license = licenses.unfree;
+    platforms = platforms.linux;
```
markuskowa, line 144:
```suggestion
    platforms = [ "x86_64-linux" ];
```
This makes the `assert` in the beginning unnecessary.

markuskowa (changes requested):

tests/turbomole/default.nix
sheepforce: added the energy of one excited state as a test condition

sheepforce:
Done with your comments :slightly_smiling_face:

markuskowa (approved):
Thanks. Looks good.

PR #10, opened 2021-04-23
MRCC 2020

sheepforce:
Addition of current MRCC binary distribution. Unfortunately the sources only build properly with Intel tooling. A terrible workaround is used for a very simple problem; the basis set files in the installation directory are required to be writable! MRCC does not actually write to it, but inherits the permissions when it copies them to the work directory and then crashes. As everything in MRCC is very hardcoded, one cannot point to another basis set directory. The solution here is ... not so nice but the best the developers and I could come up with, until this will hopefully be fixed. (Please see mrcc/default.nix to see the wrapper magic that is being used).

As for Turbomole I have added a test case, that should probably somehow be configured by cfg.nix or something?

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -151,6 +151,8 @@ let
 
       molcas2102 = callPackage ./openmolcas/v21.02.nix { };
 
+      mrcc = callPackage ./mrcc { };
+
       mt-dgemm = callPackage ./mt-dgemm { };
 
       multiwfn = callPackage ./multiwfn { };
@@ -286,6 +288,7 @@ let
         mesa-qc = callPackage ./tests/mesa { };
         molcas = callPackage ./tests/molcas { };
         #molcasUnstable = callPackage ./tests/molcas { molcas=self.molcasUnstable; };
+        mrcc = callPackage ./tests/mrcc { };
         nwchem = callPackage ./tests/nwchem { };
         psi4 = callPackage ./tests/psi4 { };
         qdng = callPackage ./tests/qdng { };
--- a/mrcc/default.nix
+++ b/mrcc/default.nix
@@ -1,0 +1,165 @@
+{ stdenv, lib, makeWrapper, requireFile, writeScript, writeTextFile, which, coreutils }:
+let
+  localRelPrefix = "mrcc";
+
+  /*
+  Creation of wrapper scripts, that replace the entry points of MRCC, namely dmrcc and dmrcc_mpi.
+  Those wrapper entrypoints create a local mrcc installation in the calling directory, that allows
+  to obtain correct permissions for the basis sets.
+  */
+  commonEntry = ''
+    # Define a cleanup function after crashes or normal termination of MRCC.
+    function cleanup {
+      rm -r $(pwd)/${localRelPrefix}
+      export PATH=$originalPATH
+      umask $originalUmask
+    }
+
+    # Look for all mrcc executables in the original installation directory.
+    mrccExes=$(find @out@/bin -type f -executable ! -path BASIS ! -path MTESTS)
+
+    # Create a local prefix for the mrcc installation and link all executables from the original
+    # location to the new local prefix.
+    mkdir -p $(pwd)/${localRelPrefix} && chown -R $USER $(pwd)/${localRelPrefix} && chmod u+rwx $(pwd)/${localRelPrefix}
+    for exe in $mrccExes; do
+      ln -s $exe $(pwd)/${localRelPrefix}/.
+    done
+
+    # To make the wrappers work again from the local prefix, replace all the original mrcc-prefix
+    # paths.
+    mrccLocalWrappers=$(find $(pwd)/${localRelPrefix}/ -type f -executable ! -name ".*-wrapped")
+    for exe in $mrccLocalWrappers; do
+      sed -i "s!@out@/bin/!$(pwd)/${localRelPrefix}/!g" $exe
+    done;
+
+    # Set a umask to make sure all files created during MRCC run have correct permissions.
+    originalUmask=$(umask)
+    umask 0023
+
+    # The Basis directory must be copied and not just symlinked. This ensures, that the basis files
+    # have correct permissions in there.
+    cp --no-preserve=all -r @out@/share/BASIS ${localRelPrefix}/.
+
+    # Save the original old path variable to restore it later and then make sure the local MRCC
+    # installation prefix is the place where all the mrcc exes are found.
+    originalPATH=$PATH
+    export PATH=$(pwd)/${localRelPrefix}:$PATH
+
+    # Create a trap if MRCC fails.
+    trap cleanup EXIT ERR
+  '';
+  dmrccEntryScript = writeScript "dmrcc" (commonEntry + ''
+      $(pwd)/${localRelPrefix}/dmrcc-original
+    '');
+  dmrccMpiEntryScript = writeScript "dmrcc_mpi" (commonEntry + ''
+      $(pwd)/${localRelPrefix}/dmrcc_mpi-original
+    '');
+
+in
+  /*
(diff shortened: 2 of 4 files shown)


mrcc/default.nix
```diff
@@ -0,0 +1,187 @@
...
+      makeWrapper
+    ];
+
+    propagatedBuildInputs = [
+      bash
+      intelmpi
+    ];
+
+    src = requireFile rec {
+      name = "mrcc.2020-02-22.binary.tar.gz";
```
markuskowa, line 85: Please use the  `version` attribute here to generate the name.

mrcc/default.nix
```diff
@@ -0,0 +1,187 @@
...
+      sha256 = "e7b4944c66b1b127f9e7b47ca973eee189a6085d4bbc8b2e46b082e57e49ad0b";
+      url = "https://www.mrcc.hu/index.php/download-mrcc/mrcc-binary/send/4-mrcc-binary/31-mrcc-2020-02-22-binary-tar";
```
markuskowa, line 87: Same as above but with `lib.replaceStrings ["."] ["-"] version`.

mrcc/default.nix
```diff
@@ -0,0 +1,187 @@
...
+let
+  localRelPrefix = "mrcc";
+
+  /*
+  Creation of wrapper scripts, that replace the entry points of MRCC, namely dmrcc and dmrcc_mpi.
+  Those wrapper entrypoints create a local mrcc installation in the calling directory, that allows
+  to obtain correct permissions for the basis sets.
+  */
+  commonEntry = ''
+    #!${bash}/bin/bash
```
markuskowa, line 17: That line shouldn't be necessary since you use `writeScript` later?

mrcc/default.nix
```diff
@@ -0,0 +1,187 @@
...
+
+      runHook postInstall
+    '';
+
+    # Manual fix of the RPATHs of the partially dynamically linked MPI programs
+    preFixup = ''
+      cd $out/bin
+      dynamicExes="dmrcc_mpi mrcc_mpi ccsd_mpi scf_mpi"
+      for exe in $dynamicExes; do
+        patchelf --set-rpath ${intelmpi}/lib:${stdenv.cc.cc.lib}/lib $exe
```
markuskowa, line 133: the `autoPatchElfHook` fails here?

tests/mrcc/MINP
```diff
@@ -0,0 +1,23 @@
+# Local CCSD(T) for a five H2 cluster with cc-pVDZ basis testing the 2018 scheme with vloose settings
+mem=50mb
+calc=LNO-CCSD(T)
+basis=cc-pVDZ
+
+localcc=2018
+lcorthr=vloose
+
+test=-5.662212025812
```
markuskowa, line 9: Does the run fail if the single point calculation does not give exactly this value?

mrcc/default.nix
```diff
@@ -0,0 +1,187 @@
...
+  stdenv.mkDerivation rec {
+    pname = "mrcc";
+    version = "2020.02.22";
+
+    nativeBuildInputs = [
+      makeWrapper
+    ];
+
+    propagatedBuildInputs = [
+      bash
```
markuskowa, line 80: Is bash really need here?

default.nix
```diff
@@ -182,6 +182,10 @@ let
 
       vmd = callPackage ./vmd {};
 
+      intelmpi = callPackage ./intelmpi { };
```
markuskowa, line 185: Please use alphabetical order here.

markuskowa (changes requested):

mrcc/default.nix
sheepforce: yes, autoPatchelf fails on some of the static executables and produces a non-0 exit code.

tests/mrcc/MINP
sheepforce: I guess it would. But I actually had never any failing test in the MRCC test suite.

sheepforce:
OK, fixed by your suggestions :slightly_smiling_face: What is left is the question how to deal with the test cases and the license/availability problem.

I would like to add Pysisyphus later but it would be similar for Psi4. Both can depend on MRCC at runtime for CC calculations, so it is an optional build input. In my own overlay I was using a configuration file, which sets the availability of such codes.

markuskowa:
The test should be fine as is. You can run it with `nix-build release.nix -A qchem.tests.mrcc`. There is no configuration required. For all derivations that use `mrcc` as input it should simply be optional. I am not sure yet if having additional config options is the best solution here. Let's have a second look when we are done with all the PRs.

Btw.: I was looking `intelmpi`. There seems to be some error:
```
$ result/bin/impi_info
result/bin/impi_info: line 8: /nix/store/i480y85mlhkqj3yafyx83ql727igxgs7-intelmpi-2019.7.217/bin/.impi_info-wrapped: No such file or directory
```

intelmpi/default.nix
```diff
@@ -0,0 +1,89 @@
+{ stdenv, lib, gfortran, fetchurl, makeWrapper, rpmextract, which
+# Runtime dependencies
+, openssh
+}:
+stdenv.mkDerivation rec {
+    pname = "intelmpi";
+    version = "2019.7.217";
```
markuskowa, line 7: Is that the latest version?

sheepforce: No, it is not. It is just the version that works for MRCC. The Intel toolings have changed significantly in 2021. There is this HPC OneAPI" thing now, which includes the compilers, MPI and some analysis programs. 

In principle I think it would be very nice to package those. MRCC is nearly 20 times slower when building the minimal version with gfortran instead of ifort and MultiWFN does not build at all. But building a stdenv around the intel tools has been attempted and does not look easy (https://github.com/NixOS/nixpkgs/issues/32434).

But here in this case it would be a very different derivation. We could also just drop IntelMPI for now. It actually runs much better in the OMP parallel version and I had never any real use for the MPI parallel version.

sheepforce:
> The test should be fine as is. You can run it with `nix-build release.nix -A qchem.tests.mrcc`. There is no configuration required. For all derivations that use `mrcc` as input it should simply be optional. I am not sure yet if having additional config options is the best solution here. Let's have a second look when we are done with all the PRs.
> 
> Btw.: I was looking `intelmpi`. There seems to be some error:
> 
> ```
> $ result/bin/impi_info
> result/bin/impi_info: line 8: /nix/store/i480y85mlhkqj3yafyx83ql727igxgs7-intelmpi-2019.7.217/bin/.impi_info-wrapped: No such file or directory
> ```

Hm, I've never used it besides MRCC. See above. If we decide to drop IntelMPI for now, I will just remove it. Otherwise I will have another look.

markuskowa:
The `bin/impi_info` is actually a good thing to have. We should keep it. It looks like a simple wrapper error.

sheepforce:
> The `bin/impi_info` is actually a good thing to have. We should keep it. It looks like a simple wrapper error.

I meant we could remove the whole IntelMPI distribution for now, it would not be too important for MRCC. But I head a look at `impi_info`. It is simply the the LD interpreter, that still points to `/lib64/ld-linux-x86-64.so.2`. Probably this would be easy to fix with `patchelf` but as with MKL in NixPkgs the license does not allow to touch the binaries. The other executables are all statically linked, so this is not an issue for them.
Do you know how to deal with such a case? I have no good idea.

markuskowa:
OK, Then I misunderstood it. If `intelmpi` is not strictly necessary then let's remove it for now. It builds clean with openmpi for example.

sheepforce:
Alright. Using it without IntelMPI for now, so only the SMP parallel version for now. Please have another look :slightly_smiling_face:

markuskowa (approved):

PR #11, opened 2021-04-23
GPAW 21.1.0

sheepforce:
Adds GPAW as python library and application. Unfortunately I have no idea how to get the checkPhase working. It requires working MPI and fails with some missing rsh. Works otherwise.

Some probably important change: i habe enabled the AVX tunings and MPI for FFTW in the overlay. GPAW requires the MPI bindings of FFTW.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -138,6 +138,8 @@ let
 
       mt-dgemm = callPackage ./mt-dgemm { };
 
+      multiwfn = callPackage ./multiwfn { };
+
       orca = callPackage ./orca { };
 
       osu-benchmark = callPackage ./osu-benchmark {
@@ -180,10 +182,10 @@ let
 
       stream-benchmark = callPackage ./stream { };
 
+      turbomole = callPackage ./turbomole {};
+
       vmd = callPackage ./vmd {};
 
-      multiwfn = callPackage ./multiwfn {};
-      turbomole = callPackage ./turbomole {};
 
 
 
--- a/default.nix
+++ b/default.nix
@@ -203,6 +203,8 @@ let
 
       libint2 = callPackage ./libint { inherit optAVX; };
 
+      libvdwxc = callPackage ./libvdwxc { };
+
       # libint configured for bagel
       # See https://github.com/evaleev/libint/wiki#bagel
       libint-bagel = callPackage ./libint { cfg = [
--- a/default.nix
+++ b/default.nix
@@ -63,6 +63,19 @@ let
         buildInputs = [ self_.gfortran ];
       });
 
+      fftw-mpi = self.fftw.overrideAttrs (oldAttrs: {
+        buildInputs = oldAttrs.buildInputs ++ [
+          self.mpi
+        ];
+
+        configureFlags = with lib.lists; oldAttrs.configureFlags ++ [
+          "--enable-mpi"
+          "MPICC=${self_.mpi}/bin/mpicc"
+          "MPIFC=${self_.mpi}/bin/mpif90"
+          "MPIF90=${self_.mpi}/bin/mpif90"
+        ];
+      });
+
       # For molcas and chemps2
       hdf5-full = self_.hdf5.override {
         cpp = true;
@@ -113,6 +126,8 @@ let
       ergoscf = callPackage ./ergoscf { };
 
       gaussview = callPackage ./gaussview { };
+
+      gpaw = super.python3.pkgs.toPythonApplication self.python3.pkgs.gpaw;
 
       nwchem = callPackage ./nwchem { blas=self.blas-i8; lapack=self.lapack-i8; };
 
--- a/gpaw/default.nix
+++ b/gpaw/default.nix
@@ -1,0 +1,126 @@
+{ fetchFromGitLab, buildPythonPackage, lib, writeTextFile, makeWrapper, fetchurl, blas, lapack,
+  scalapack, mpi, fftw, libxc, libvdwxc, which, ase, numpy, scipy
+}:
+assert
+  lib.asserts.assertMsg
+  (!blas.isILP64)
+  "A 32 bit integer implementation of BLAS is required.";
+
+assert
+  lib.asserts.assertMsg
+  (!lapack.isILP64)
+  "A 32 bit integer implementation of LAPACK is required.";
+
+let
+  gpawConfig = writeTextFile {
+    name = "siteconfig.py";
+    text = ''
+      # Compiler
+      compiler = 'gcc'
+      mpicompiler = '${mpi}/bin/mpicc'
+      mpilinker = '${mpi}/bin/mpicc'
+
+      # BLAS
+      libraries += ['blas']
+      library_dirs += ['${blas}/lib']
+
+      # FFTW
+      fftw = True
+      if fftw:
+        libraries += ['fftw3']
+
+      scalapack = True
+      if scalapack:
+        libraries += ['scalapack']
+
+      # LibXC
+      libxc = True
+      if libxc:
+        xc = '${libxc}/'
+        include_dirs += [xc + 'include']
+        library_dirs += [xc + 'lib/']
+        extra_link_args += ['-Wl,-rpath={xc}/lib'.format(xc=xc)]
+        if 'xc' not in libraries:
+          libraries.append('xc')
+
+      # LibVDWXC
+      libvdwxc = True
+      if libvdwxc:
+        vdwxc = '${libvdwxc}/'
+        extra_link_args += ['-Wl,-rpath=%s/lib' % vdwxc]
+        library_dirs += ['%s/lib' % vdwxc]
+        include_dirs += ['%s/include' % vdwxc]
+        libraries += ['vdwxc']
+    '';
+  };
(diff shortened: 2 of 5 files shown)


default.nix
```diff
@@ -60,7 +60,19 @@ let
                 else super.lapack.override { isILP64 = true; lapackProvider = super.openblas; };
 
       fftw = self_.fftw.overrideDerivation ( oldAttrs: {
```
markuskowa, line 62: Let's make this  a separate attribute: `fftw-mpi`.  Compiling mpi into the standard `fftw` means that all applications will depend on MPI even if they do not use it.

I am note sure why I used `overrideDerivation` here. It should be `overrideAttributes`.

gpaw/default.nix
```diff
@@ -0,0 +1,143 @@
+{ fetchFromGitLab, buildPythonPackage, lib, writeTextFile, makeWrapper, fetchurl
+, blas
```
markuskowa, line 2: Please use uniform formatting here.

gpaw/default.nix
```diff
@@ -0,0 +1,143 @@
...
+
+    src = fetchFromGitLab  {
+      owner = "gpaw";
+      repo = pname;
+      rev = version;
+      sha256 = "0xws39yq5d1ih4zj31bdd0h64afs6x9j30yrrycxf3pnc5h9wbr0";
+    };
+
+    preConfigure = ''
+      export PATH=$PATH:${mpi}/bin
```
markuskowa, line 106: Shouldn't this be in the path automatically?

gpaw/default.nix
```diff
@@ -0,0 +1,143 @@
...
+      mkdir -p $out/share/gpaw && cd $out/share/gpaw
+      cp ${pawDataSets} gpaw-setups.tar.gz
+      tar -xvf $out/share/gpaw/gpaw-setups.tar.gz
+      rm gpaw-setups.tar.gz
+      cd $currDir
+    '';
+
+    doCheck = false;
+    preCheck = ''
+      # export PATH=$PATH:${mpi}/bin
```
markuskowa, line 120: this line can be removed.

gpaw/default.nix
```diff
@@ -0,0 +1,143 @@
...
+    postPatch = ''
+      cp ${gpawConfig} siteconfig.py
+    '';
+
+    postFixup = ''
+      for exe in $out/bin/*; do
+        wrapProgram $exe \
+          --set "GPAW_SETUP_PATH" "$out/share/gpaw/gpaw-setups-${setupVersion}"
+      done
+      ln -s ${mpi}/bin/mpiexec $out/bin/.
```
markuskowa, line 133: I have done something like this for CP2K but I am not convinced we should keep linking `mpirun` into `$out/bin` as this could lead to collisions. Let's remove it for now and look for a better (nix compatible) way to do this.

gpaw/default.nix
```diff
@@ -0,0 +1,143 @@
...
+      ln -s ${mpi}/bin/mpirun $out/bin/.
+    '';
+
+    meta = with lib; {
+      description = "DFT and beyond within the projector-augmented wave method";
+      license = licenses.gpl3;
```
markuskowa, line 139: `gplN` is deprecated. It needs to be specified now as `gplNOnly` or `gplNPlus`.

libvdwxc/default.nix
```diff
@@ -0,0 +1,55 @@
+{ stdenv, lib, fetchFromGitLab, gfortran, autoreconfHook
+, fftw
+, mpi
+}:
+stdenv.mkDerivation rec {
+  pname = "libvdwxc";
+  # Stable version has non-working MPI detection.
+  version = "92f4910c6ac88e111db2fb3a518089d0510c53b0";
```
markuskowa, line 8: Please us a date stamp for version instead of a hash value.

libvdwxc/default.nix
```diff
@@ -0,0 +1,55 @@
...
+    sha256= "1c7pjrvifncbdyngs2bv185imxbcbq64nka8gshhp8n2ns6fids6";
+  };
+
+  hardeningDisable = [
+    "format"
+  ];
+
+  meta = with lib; {
+    description = "Portable C library of density functionals with van der Waals interactions for density functional theory";
+    license = with licenses; [ lgpl3 bsd3 ];
```
markuskowa, line 51: Same as above for `gpl3`

default.nix
```diff
@@ -182,6 +194,7 @@ let
 
       vmd = callPackage ./vmd {};
 
+      gpaw = super.python3.pkgs.toPythonApplication self.python3.pkgs.gpaw;
```
markuskowa, line 197: Please place the attributes in alphabetical order in the respective sub sections.

markuskowa (changes requested):

gpaw/default.nix
sheepforce: removed linking of MPI. On a first thought i have no idea how to provide the MPI executables nicely

default.nix
sheepforce: Added a separate attribute `fftw-mpi`. Interestingly the original `fftw` derivation does not seem to define `overrideAttributes`. It is simply not available.

sheepforce:
with the new `fftw-mpi` GPAW does not find FFTW anymore. I will try to fix this and then we should be good to go :slightly_smiling_face:

sheepforce:
Done with all your comments. Please have another look :slightly_smiling_face:

default.nix
markuskowa: Does it work again if you fix the the above line to:
```
fftw = self_.fftw.overrideAttributes ( oldAttrs: {
```
?

gpaw/default.nix
markuskowa: There are several ways to deal with it:
1. On an HPC cluster with properly set up PMIX one can just use `srun` instead of `mpirun`. 
2. Setting up the job with e.g. `nix-shell -p gpaw mpi` also works (little more typing though but a clean solution). 
3. We could use the `passthru` attribute:
```
passthru.mpi = mpi;
```
Then you can compose your environment like this: `nix-shell -p gpaw gpaw.passthru.mpi` (even more typing but a very clean solution).

default.nix
```diff
@@ -63,6 +63,19 @@ let
         buildInputs = [ self_.gfortran ];
       });
 
+      fftw-mpi = self.fftw.overrideDerivation (oldAttrs: {
```
markuskowa, line 66: I think that should be `overrideAttributes` here (also for the override of `fftw`. I probably made a mistake here using `overrideDerivation` in the first place).

markuskowa:
Can it be that the download URL for `libvdwxc` is outdated?
```
these derivations will be built:
  /nix/store/d794zcd8vszyqqgbnpwdrf6x2frmcl6g-source.drv
  /nix/store/rl3p0vhk832kyj1ix3k74swmi1vkrgzl-libvdwxc-24.02.2020.drv
building '/nix/store/d794zcd8vszyqqgbnpwdrf6x2frmcl6g-source.drv'...

trying https://gitlab.com/api/v4/projects/libvdwxc%2Flibvdwxc/repository/archive.tar.gz?sha=24.02.2020
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404
error: cannot download source from any mirror

```

default.nix
sheepforce: Actually FFTW does not export overrideAttributes. I was also a little bit surprised but I have tried after your first comment.

sheepforce:
> Can it be that the download URL for `libvdwxc` is outdated?
> 
> ```
> these derivations will be built:
>   /nix/store/d794zcd8vszyqqgbnpwdrf6x2frmcl6g-source.drv
>   /nix/store/rl3p0vhk832kyj1ix3k74swmi1vkrgzl-libvdwxc-24.02.2020.drv
> building '/nix/store/d794zcd8vszyqqgbnpwdrf6x2frmcl6g-source.drv'...
> 
> trying https://gitlab.com/api/v4/projects/libvdwxc%2Flibvdwxc/repository/archive.tar.gz?sha=24.02.2020
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
>   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
> curl: (22) The requested URL returned error: 404
> error: cannot download source from any mirror
> ```

Ah, I made a mistake. I've been using the version in the src field and changed as you requested to the date. This of course broke the URL. But I didn't notice as I had the archive with the correct sha256 hash already in store, so it still built. I will fix it.

default.nix
markuskowa: My bad, I meant `overrideAttrs`. I fixed it on master. Just do a rebase.

sheepforce: Done

sheepforce:
Fixed the libvdwxc URL.

PR #12, opened 2021-04-23
Multiwfn 3.7

sheepforce:
Adds Multiwfn to the package set. It's a Fortran program but absolutely only builds with Intel's ifort. The author is aware and not willing to fix it. As long as the Intel compilers don't make it into nixpkgs the binary is the only option.

### DIFF
--- a/multiwfn/default.nix
+++ b/multiwfn/default.nix
@@ -1,0 +1,49 @@
+{ stdenv, lib, makeWrapper, requireFile, autoPatchelfHook, unzip,
+  writeScriptBin, fetchurl, glibc, xorg, libGL, motif
+}:
+stdenv.mkDerivation rec {
+    pname = "multiwfn";
+    version = "3.7";
+
+    nativeBuildInputs = [
+      makeWrapper
+      autoPatchelfHook
+      unzip
+    ];
+
+    buildInputs = [
+      xorg.libX11
+      libGL
+      motif
+    ];
+
+    # Only builds with Intel tooling. ifort is not available in Nix, therefore the binary package.
+    src = fetchurl {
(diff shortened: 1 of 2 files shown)


multiwfn/default.nix
```diff
@@ -0,0 +1,54 @@
+{ stdenv, lib, makeWrapper, requireFile, autoPatchelfHook, unzip, writeScriptBin, fetchurl
```
markuskowa, line 1: regarding formatting: both versions, single or multiple packages per line, are OK. However, I would prefer to stick to one version.

multiwfn/default.nix
```diff
@@ -0,0 +1,54 @@
...
+, bash
```
markuskowa, line 2: `bash` is not used.

multiwfn/default.nix
```diff
@@ -0,0 +1,54 @@
...
+    version = "3.7";
+
+    nativeBuildInputs = [
+      makeWrapper
+      autoPatchelfHook
+      unzip
+    ];
+
+    buildInputs = [
+      glibc
```
markuskowa, line 19: Do we really need `glibc` in the build inputs for the package to work?

markuskowa (changes requested):

markuskowa:
One more comment. I just noticed the format of the commit messages. Can you please use the standard nixpkgs upstream convention: `package: init at X.Y` or `package: fix xyz`

multiwfn/default.nix
sheepforce: Actually not. Have removed it

markuskowa (approved):
Thanks. Looks good now.

PR #15, opened 2021-04-27
Grimme Programs

sheepforce:
This adds the following programs:
  - Cefine - non-interactive wrapper interface to Turbomoles interactive `define`. Used by the other Grimme software
  - XTB 6.4.0 - a semiempirical and very fast tight binding code
  - CREST 2.11 with fixes - conformere sampling tool, utilising XTB as QC engine

I also fixed an environment issue with turbomole, where Nix sets `$PAGER` to `more -R`. Turbomole uses a `which $PAGER`, which obviously fails, so for turbomole I set `PAGER=more`.

There are a few issues with the upstream distributions: while in nixos-20.09 XTB formally works with `blas`, it does not in `nixpkgs-unstable`. The convergence behaviour with OpenBLAS and MKL is different already in nixos-20.09 (and much better and faster for MKL) and no convergence at all occurs with OpenBLAS in nixpkgs-unstable, while it is much slower in nixos-20.09. Therefore I simply pass MKL to XTB. There are also issues with threading in XTB when using OpenBLAS. The Grimme group likes Intel tooling very much ... The same is true for CREST.

XTB also has IO based interfaces to ORCA and Turbomole. The ORCA interface works without any problems, while the Turbomole interface is based around Cefine. While Cefine itself works perfectly and produces valid inputs that nicely run with Turbomole, XTB fails to execute the workflow correctly. This is also true when I try this outside of Nix, so this does not seem to be an issue with Nix-wrapping. Therefore, for the moment I pass turbomole and cefine as `null` to `xtb`. 

Now we start hitting the issue with non-free codes, that are dependencies of formally free codes (Turbomole and ORCA being optional dependencies of XTB). Do you have a good idea how we could more dynamically influence the activation of such proprietary codes in dependencies?


cefine/default.nix
```diff
@@ -0,0 +1,56 @@
...
+    postFixup = ''
+      wrapProgram $out/bin/cefine \
+        --prefix PATH : "${binSearchPath}"
+    '';
+
+    meta = with lib; {
+      description = "Non-interactive command-line wrapper around turbomoles define";
+      license = licenses.lgpl3Only;
+      homepage = "https://github.com/grimme-lab/cefine";
+      platforms = platforms.unix;
```
markuskowa, line 54: You have used `platforms.unix` in other places. Is this intentional? E.g. are you using these programs also on a MAC? If not I would recommend setting it to `platforms.linux`.

xtb/default.nix
```diff
@@ -0,0 +1,50 @@
+{ stdenv, lib, gfortran, fetchFromGitHub, cmake, makeWrapper, mkl,
+  turbomole ? null, orca ? null, cefine ? null
+}:
+let
+  # Syscalls that XTB might make.
+  binSearchPath = with lib.lists; lib.strings.makeSearchPath "bin" ([ ]
```
markuskowa, line 6: is with `with lib.lists` really needed here? Wouldn't `with lib;` not more compact here?

markuskowa (changes requested):

cefine/default.nix
```diff
@@ -0,0 +1,56 @@
...
+  stdenv.mkDerivation rec {
+    pname = "cefine";
+    version = "2.23";
+
+    nativeBuildInputs = [
+      gfortran
+      makeWrapper
+    ];
+
+    propagatedBuildInputs = [
```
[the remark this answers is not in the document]
sheepforce: I like the uniform formatting more, but I have changed single-element lists to a single line :slightly_smiling_face:

cefine/default.nix
sheepforce: I am not using them on MAC, no. In principle I see no reason why they shouldn't work on a MAC, but I switched to Linux to be safe.

sheepforce:
All done.

cefine/default.nix
markuskowa: There is no clear policy about it in upstream `nixpkgs` but I usually go with "If I can't test it on another platform than Linux than I assume it only works on Linux".

PR #16, opened 2021-04-28
gdma: init at 2.3.3

sheepforce:
Adds GDMA, which performs distributed multipole analyses. Super convenient for electronic embedding calculations and used in our own software.


gdma/default.nix
```diff
@@ -0,0 +1,41 @@
+{ stdenv, lib, gfortran, fetchgit, python3, git, writeTextFile }:
```
markuskowa, line 1: writeTextFile is unused

PR #17, opened 2021-04-28
travis: init at 03.06.2020

sheepforce:
Adds travis for many MD analyses methods. Especially useful with CP2K and LibVori to calculate vibrational spectra from AIMD.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -193,6 +193,8 @@ let
 
       stream-benchmark = callPackage ./stream { };
 
+      travis-analyzer = callPackage ./travis-analyzer { };
+
       turbomole = callPackage ./turbomole {};
 
       vmd = callPackage ./vmd {};
--- a/travis-analyzer/default.nix
+++ b/travis-analyzer/default.nix
@@ -1,0 +1,24 @@
+{ stdenv, lib }:
+stdenv.mkDerivation rec {
+    pname = "travis-analyzer";
+    version = "03Jun2020";
+
+    src = fetchTarball  {
+      url = "http://www.travis-analyzer.de/files/travis-src-200504-hf2.tar.gz";
+      sha256= "1244w0sqb976c038hfkxvj4iym9nfbzhnisscrzx6vdr7q1g8pxh";
+    };
+
+    dontConfigure = true;
+    enableParallelBuilding = true;
+    installPhase = ''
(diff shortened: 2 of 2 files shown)


travis/default.nix
```diff
@@ -0,0 +1,24 @@
...
+    enableParallelBuilding = true;
+    installPhase = ''
+      mkdir -p $out/bin
+      cp exe/travis $out/bin/.
+    '';
+
+    meta = with lib; {
+      description = "Molecular dynamics trajectory analyzer and visualizer";
+      license = licenses.lgpl3Only;
+      homepage = "https://www.chemie.uni-bonn.de/pctc/mulliken-center/grimme/software/xtb";
```
markuskowa, line 21: The link is dead.

default.nix
```diff
@@ -193,6 +193,8 @@ let
 
       stream-benchmark = callPackage ./stream { };
 
+      travis = callPackage ./travis { };
```
markuskowa, line 196: The name is a bit unfortunate. There is also Travis CI in upstream nixpkgs (which is also called travis). Right now we use a subset which avoids name collisions but I would prefer to keep names separate from upstream package names. How about `travis-analyzer` or `travis-md`?

markuskowa (changes requested):

travis/default.nix
sheepforce: It was also plainly wrong. :disappointed:

default.nix
sheepforce: Switched to `travis-analyzer`.

markuskowa (approved):

travis-analyzer/default.nix
```diff
@@ -0,0 +1,24 @@
+{ stdenv, lib }:
+stdenv.mkDerivation rec {
+    pname = "travis-analyzer";
+    version = "03Jun2020";
+
+    src = fetchTarball  {
```
markuskowa, line 6: Btw.: `fetchTarball` is a builtin in function and should not be used in derivations since these will fail in `strict mode` (i.e. Hydra fails to evaluate the expression). `fetchTarball` is not strict since it would work even without the hash.
fixed in f27e9d60864368e8ad5d8d9f48c640a0683aa9af

PR #18, opened 2021-04-28
tinker: init at 8.8.3

sheepforce:
Adds the Tinker MM engine. Only implementation of AMOEBA polarisable force fields. Used by LICHEM QM/MM engine.


tinker/default.nix
```diff
@@ -0,0 +1,40 @@
+{ stdenv, lib, gfortran, cmake, writeTextFile, fftw, pkgconfig }:
```
markuskowa, line 1: writeTextFile is unused.

PR #21, opened 2021-04-28
wfoverlap: init at 24.08.2020

sheepforce:
Standalone version of the WFOverlap program, that is included in the sharc distribution or downloadable at [https://sharc-md.org/?page_id=309](https://sharc-md.org/?page_id=309). It requires some tweaks to build without MKL but works flawless then. 

The license is a little bit curious. The entire SHARC repo is under a GPL and the WFOverlap source AND binary is included there. So I would assume a GPL is fine. But in contrast there is also the registration page (see above), with a completely different non-free license.

### DIFF
--- a/wfoverlap/default.nix
+++ b/wfoverlap/default.nix
@@ -1,0 +1,51 @@
+{ stdenv, lib, fetchFromGitHub, gfortran, blas-i8, lapack-i8 }:
+assert
+  lib.asserts.assertMsg
+  (blas-i8.isILP64 || blas-i8.passthru.implementation == "mkl")
+  "64 bit integer BLAS implementation required.";
+
+assert
+  lib.asserts.assertMsg
+  (lapack-i8.isILP64)
+  "64 bit integer LAPACK implementation required.";
+
+stdenv.mkDerivation rec {
+    pname = "wfoverlap";
+    version = "24.08.2020";
+
+    src =
+      let
+        repo = fetchFromGitHub {
+          owner = "sharc-md";
+          repo = "sharc";
+          rev = "d943ec7aff0fb6c81f61d3c057b0921d053e9e20";
+          sha256 = "1a9frnxvm1jg4cv3jd4lm3q8m7igyc7fsp3baydfjkvk0b4ss9bc";
+        };
+      in "${repo}/wfoverlap/source";
+
+    nativeBuildInputs = [ gfortran ];
+
+    buildInputs = [
+      gfortran
+      blas-i8
+      lapack-i8
+    ];
+
+    patches = [ ./Makefile.patch ];
+
+    dontConfigure = true;
+
+    hardeningDisable = [ "format" ];
+
+    installPhase = ''
+      mkdir -p $out/bin
+      cp wfoverlap.x $out/bin/.
+    '';
+
+    meta = with lib; {
+      description = "Efficient calculation of wavefunction overlaps";
+      license = licenses.gpl3Only;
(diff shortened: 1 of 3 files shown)


wfoverlap/default.nix
```diff
@@ -0,0 +1,70 @@
...
+    ];
+
+    patches = [
+      ./patches/Makefile.patch
+    ];
+
+    postPatch = ''
+       export blasLibs="${blasLibs}"
+       export lapackLibs="-llapack"
+       ls -lah ./Makefile
```
markuskowa, line 48: This was for debugging purposes?

wfoverlap/default.nix
```diff
@@ -0,0 +1,70 @@
+{ stdenv, lib, fetchFromGitHub, gfortran, makeWrapper, blas-i8, lapack }:
+assert
+  lib.asserts.assertMsg
+  (blas-i8.isILP64 || blas-i8.passthru.implementation == "mkl")
+  "64 bit integer BLAS implementation required.";
+
+let
+  useMKL64 = blas-i8.passthru.implementation == "mkl";
```
markuskowa, line 8: I gave it a quick try and removed all of MKL distinction and could build it without problem against MKL (`nix-build release.nix -A qchem-mkl.wfoverlap --arg buildVariants true`). I think we can remove all of that code?

wfoverlap/default.nix
```diff
@@ -0,0 +1,70 @@
+{ stdenv, lib, fetchFromGitHub, gfortran, makeWrapper, blas-i8, lapack }:
```
markuskowa, line 1: Same comment here. I think both should be `*-i8` here. Also the convention in `nixpkgs` would tend to write `blas` and `lapack` in the input parameters and then set callPackage:
```
wfoverlap = callPackage ./wfoverlap { blas=self.blas-i8; lapack=lapack-i8; };
```

wfoverlap/default.nix
```diff
@@ -0,0 +1,70 @@
...
+          --subst-var blasLibs \
+          --subst-var lapackLibs
+    '';
+
+    dontConfigure = true;
+
+    hardeningDisable = [ "format" ];
+
+    installPhase = ''
+      cat Makefile
```
markuskowa, line 59: this line can be removed?

wfoverlap/default.nix
```diff
@@ -0,0 +1,70 @@
...
+    ];
+
+    buildInputs = [
+      gfortran
+      (if useMKL64 then mkl else blas-i8)
+      lapack
+    ];
+
+    patches = [
+      ./patches/Makefile.patch
```
markuskowa, line 42: no need for a separate patch sub directory.

markuskowa (changes requested):

wfoverlap/default.nix
sheepforce: I've used the variable substitute mechanism originally. No longer required.

wfoverlap/default.nix
sheepforce: You are right. When switching to lapack-i8 it builds flawless. Changed.

sheepforce:
All done!

wfoverlap/default.nix
```diff
@@ -0,0 +1,54 @@
+{ stdenv, lib, fetchFromGitHub, gfortran, makeWrapper, blas-i8, lapack-i8 }:
```
markuskowa, line 1: makeWrapper is not used.

markuskowa (approved):

PR #22, opened 2021-04-28
cfour: init at 2.1

sheepforce:
This adds the public CFour 2.1, which is available after registration but free of charge. Some specialties of CFour: with GCC 10 and the current GFortran from nixpkgs master it produces segfaults, which did not happen with stdenv from nixos-20.09. This seems to be a known bug in GCC 10 for some very special functions. It is necessary to disable the stack protector and string optimisations to work around the segfaults.

Furthermore CFour works in principle with OpenMPI. It is just not called directly but as with ORCA by the entry point executable `xcfour`. Instead, they are trying to obtain the number of MPI threads from an environment variable, see also the [cfour manual](http://slater.chemie.uni-mainz.de/cfour/index.php?n=Main.Examples). I try this here but `xcfour` does not see the environment variable and instead of `mpirun -np 4 xinitenv` xcfour executes `mpirun -np xinitevn`, which obviously fails. I don't see why it is not passed properly. I am open for suggestions. For the moment it works well without MPI.


cfour/default.nix
```diff
@@ -0,0 +1,66 @@
+{ stdenv, lib, makeWrapper, gfortran, requireFile, writeTextFile, python3, blas-i8, lapack,
```
markuskowa, line 1: I am surprised that mixing `blas-i8` and `lapack` works here since they are API incompatible (ILP64 vs LP64).

cfour/default.nix
```diff
@@ -0,0 +1,66 @@
...
+# Make sure to have a correct version of BLAS.
+assert
+  lib.asserts.assertMsg
+  (blas-i8.isILP64)
+  "A 64 bit integer implementation of BLAS is required.";
+
+# Only works with OpenMPI
+assert
+  lib.asserts.assertMsg
+  (mpi == null || mpi.pname == "openmpi")
```
markuskowa, line 13: It won't work with MPICH?

cfour/default.nix
```diff
@@ -0,0 +1,66 @@
...
+  "If MPI is to be used for CFour, use OpenMPI!";
+let
+  useMPI = mpi != null;
+  useMKL64 = blas-i8.passthru.implementation == "mkl";
```
markuskowa, line 17: I wonder if that should be done via `useMKL64`.  The blas/lapack switching mechanism in nixpkgs should make it easier by using overrides like this (https://nixos.org/manual/nixpkgs/stable/#sec-overlays-alternatives-blas-lapack):
```
callPackage ./cfour {
    blas = super.blas.override { blasProvider = super.mkl; };
    lapack = super.lapack.override { lapackProvider = super.mkl; };
};
```
That would save you a lot of code in the derivation where you make the distinctions between MKL/non-MKL.

cfour/default.nix
```diff
@@ -0,0 +1,66 @@
...
+    postFixup = ''
+      wrapProgram $out/bin/xcfour \
+        --set-default CFOUR_NUM_CORES 1
+    '';
+
+    meta = with lib; {
+      description = "Specialist coupled cluster software.";
+      homepage = "http://slater.chemie.uni-mainz.de/cfour/index.php";
+      license = licenses.unfree;
+      platforms = platforms.unix;
```
markuskowa, line 64: unix or linux?

cfour/default.nix
```diff
@@ -0,0 +1,66 @@
...
+  (blas-i8.isILP64)
+  "A 64 bit integer implementation of BLAS is required.";
+
+# Only works with OpenMPI
+assert
+  lib.asserts.assertMsg
+  (mpi == null || mpi.pname == "openmpi")
+  "If MPI is to be used for CFour, use OpenMPI!";
+let
+  useMPI = mpi != null;
```
markuskowa, line 16: Is there a legitimate use-case for using it without MPI? Would it make sense to remove the conditional and just build it always with MPI?

cfour/default.nix
```diff
@@ -0,0 +1,66 @@
...
+
+    buildInputs = [
+      gfortran
+      (if useMKL64 then mkl else blas-i8)
+      lapack
+    ];
+
+    propagatedBuildInputs = lib.lists.optional useMPI mpi;
+
+    src = requireFile rec {
```
markuskowa, line 34: The `rec` is not needed here.

tests/cfour/default.nix
```diff
@@ -0,0 +1,22 @@
...
+    ./GENBAS
+  ];
+
+  outFile = [ "cfour.out" ];
+
+  OMP_NUM_THREADS=2;
+
+  testScript = ''
+    @test "Run CFour" {
+      xcfour > cfour.out
```
markuskowa, line 19: In such tests it is always nice to have check the final energy. That we have a confirmation that it not only runs but always  yields the same value.

markuskowa (changes requested):

cfour/default.nix
sheepforce: The mixture of the i-8 and i-4 interfaces still produce a lot of headache for me. My understanding from what usually happens, is that for BLAS the API often allows to select i-8 or i-4, while LAPACK is not influenced by this. As far as I could see, this it what happens for OpenBLAS, when you switch to the 64 bit integers. At least it was when I did it outside of Nix. I have no idea how deep the LAPACK overrides in NixPkgs go. It is a version that I commonly saw: NWChem with i-8 OpenBLAS can still us i-4 LAPACK, the same for OpenMolcas, where actually ScaLAPACK is involved, which is more or less always i-4. There is a port of ScaLAPACK for i-8, which requires MPI with i-8 (which breaks all packages, for which I've ever tried) or Intel's ScaLAPACK, which seems to be the only ScaLAPACK that supports a working i-8. I tried for a long time with Molcas (see [here](https://gitlab.com/Molcas/OpenMolcas/-/issues/159)). 
In the cases of NWChem and Molcas, also the underlying GlobalArrays allows to mix i-8 BLAS with i-4 LAPACK (when I build against OpenBLAS with INTERFACE=64 for example) and works fine. It does definitely not when using both in i-8 versions.

cfour/default.nix
sheepforce: It won't. The configure script allows to explicitly select LAM, IntelMPI and OpenMPI and nothing else. I guess that IntelMPI and MPICH are quite similar and could work. But I guess it is an unsupported configuration.

cfour/default.nix
sheepforce: Yes, this actually a very convenient thing to have and I would like to use it more often. I enabled it only in the places, where this mechanism unfortunately fails. It seems common, that QC codes try to detect if they link against MKL and use some special flags then. The NixPkgs mechanism does not seem to be completely transparent and e.g. CFour figures out, that it is not a generic `-lblas` that it is seeing but actually MKL under the hood. It tries to use MKL flags then and fails. Therefore, I usually unwrap MKL in those cases. They are actually (and unfortunately) by intention.

cfour/default.nix
sheepforce: Sry, forgot it here. Switched to Linux.

cfour/default.nix
sheepforce: The OMP parallel version is also working nice. We could always build with MPI, but i haven't solved the problem mentioned in the original PR message with the number of cores yet.

tests/cfour/default.nix
sheepforce: Done!

cfour/default.nix
markuskowa: I think both should be the same API. When you at the NWChem derivation it uses `blas-i8` and `lapack-i8`, which works. Whenever Fortran  defines a bare integer it will depend on the compiler flags if it is i4 or i8. In BLAS and LAPACK this is used to define matrix/vector sizes in the API. LAPACK will hand matrices down to BLAS. If both use different integer sizes I would expect it to crash. I would say that they should always be the same, which should be enforced by an `assert`. Everything should be complied with the same convention: application, LAPACK, BLAS.
Btw. the SCALAPACK reference implementation only works with I4. I never managed to compile it with I8. However, Intel provides a version with I8.
Regarding MOLCAS and GA: I have given up for now compiling it with MPI (I guess I ran exactly into the problem that you are referencing).

cfour/default.nix
markuskowa: OK, let's keep the conditional.

cfour/default.nix
markuskowa: I wonder why the developers make that distinction. MPI is a very well standardized interface. IntelMPI uses the MPICH code base and is nearly identical. We could drop this assert since OpenMPI is the default anyway and it would allow to do test builds with MPICH.

cfour/default.nix
markuskowa: > It seems common, that QC codes try to detect if they link against MKL and use some special flags then. 

Unfortunately Molcas does the same thing.

> The NixPkgs mechanism does not seem to be completely transparent and e.g. CFour figures out, that it is not a generic -lblas that it is seeing but actually MKL under the hood.

You can always look for MKL specific symbols/functions in the library to figure out with implementation it is. The `nixpkgs` switching mechanism does not try to hide it but rather provides the linker flags in a discrete manner.

cfour/default.nix
sheepforce: I have removed this assert. Having MPICH builds certainly makes sense.

.gitignore
```diff
@@ -0,0 +1,3 @@
+tests.nix
+shell.nix
```
markuskowa, line 2: Please remove shell.nix and tests.nix

cfour/default.nix
```diff
@@ -0,0 +1,64 @@
...
+       ${if useMPI then "--with-exenodes=\"${mpi}/bin/mpiexec -np \$1\"" else ""}
+       FCFLAGS="-fno-optimize-strlen -fno-stack-protector"
+       CFLAGS="-fno-optimize-strlen -fno-stack-protector"
+       CXXFLAGS="-fno-optimize-strlen -fno-stack-protector"
+      )
+    '';
+
+    hardeningDisable = [ "format" ];
+
+    # postFixup = ''
```
markuskowa, line 53: Please remove it form the code if it is not needed anymore.

.gitignore
sheepforce: Sorry, you just saw my work in progress. This was never meant to be there. :slightly_smiling_face:

sheepforce:
Was finally able to get it running correctly with MPI. It was a just a problem with escapes in Nix-strings. It also actually recognises `mpich` as an option, so I removed the OpenMPI assert. 

I was also able to build it with MVAPICH2. Is there a reason why you've put it in the legacy section of the overlay? On our OmniPath cluster MVAPICH2 is significantly faster than OpenMPI, especially for CP2K, where OpenMPI also produces memory leaks in our configuration.

markuskowa:
> I was also able to build it with MVAPICH2. Is there a reason why you've put it in the legacy section of the overlay?

"legacy version" only refers to `libxc4` here. 
I haven't used MVAPICH that much since we rarely use Infiniband. I think the derivation could use a little clean up to be able to build build different variants with the flip of a switch. MVAPICH is one of the cases which should be moved upstream eventually.

markuskowa (approved):

PR #23, opened 2021-04-28
dalton: init at 2020.0

sheepforce:
Adds the Dalton program, which is a specialist in calculating exotic properties.


dalton/default.nix
```diff
@@ -0,0 +1,93 @@
...
+  */
+  postInstall = ''
+    mkdir -p $out/bin $out/share/dalton
+    for exe in dalton dalton.x; do
+      mv $exe $out/bin/.
+    done
+
+    for dir in basis tools; do
+      mv $dir $out/share/dalton/.
+    done
```
markuskowa, line 66: Now things in dalton/* and share/* are duplicated. Is that intended or is just a `rm -r $out/dalton` missing here?

sheepforce: You are right. I simply missed it. Removed the redundancy.

Issue #29, opened 2021-04-30
PySCF tests fail


sheepforce:
PySCF fails to build as its tests don't pass. It raises an assertion error on an AM1 test case, which is not implemented. It almost looks as it is not even possible for this test to pass?

[pyscf_err.txt](https://github.com/markuskowa/NixOS-QChem/files/6406070/pyscf_err.txt)

markuskowa:
`pyscf` builds fine for me against the `nixpkgs-unstable` branch.

sheepforce:
That's odd. I've tried with `nixpkgs-unstable` (commit 9775b39fd45a61c2a53aa6a90485fac8f87ce7c6) and `master` and both fail with the same error :thinking: I am not fluent in python but the error is in the sources and  triggered on module import of AM1 ([source line](https://github.com/pyscf/pyscf/blob/7be5e015b2b40181755c71d888449db936604660/pyscf/semiempirical/am1.py#L11)).
I've also tried to update PySCF ([this branch](https://github.com/sheepforce/NixOS-QChem/tree/pyscf)) but still get the same error.

markuskowa:
That is strange `nixpkgs-unstable` aebebb5 still builds just fine for me. Do run the build on NixOS or non-NixOS? Do you have sandboxing turned on?

sheepforce:
I am building on NixOS 20.09 and sandboxing is turned on.

sheepforce:
Tested with `nix-build --cores 1 --max-jobs 1 release.nix --arg config '{ allowEnv = false; }' --arg nixpkgs 'builtins.fetchTarball https://github.com/NixOS/nixpkgs/archive/1c16013bd6e94da748b41cc123c6b509a23eb440.tar.gz' -A qchem.python3.pyscf` against commit `c112eae880ff26f82dc8e4c36c7d3bb647bf7540`. Gives the same AM1 error as previously. (tried with `--cores and `--max-jobs` just to be sure that it is not a problem of parallel builds).

sheepforce:
Managed to set up a hydra. It also fails there, again with a different error (and the processors are old Haswells without AVX512 ...). This times with
```
/build/source/pyscf/lib/misc.py:47: H5pyDeprecationWarning: Using default_file_mode other than 'r' is deprecated. Pass the mode to h5py.File() instead.
  h5py.get_config().default_file_mode = 'a'
Traceback (most recent call last):
  File "nix_run_setup", line 8, in <module>
[...]

PR #31, opened 2021-04-30
pysisyphus: init at 0.7rc1

sheepforce:
This adds Pysisyphus for geometry optimisaiton methods. It implements optimisations with wavefunction root tracking, chain of state methods, AFIR and more methods for transition, ground and excited state optimisations. Pysisyphus wraps many other QC programs to obtain forces and is somehow difficult to setup without nix.

As PySCF currently fails to build for my, this PR is blocked by #29 .

This and the GAMESS-US PR #28 are the last package additions from my list. :slightly_smiling_face: :partying_face:

### DIFF
--- a/pkgs/apps/nwchem/default.nix
+++ b/pkgs/apps/nwchem/default.nix
@@ -3,7 +3,7 @@
 , automake, autoconf, libtool, makeWrapper
 } :
 
-assert blas.isILP64 == blas.isILP64;
+assert blas.isILP64 == lapack.isILP64;
 
 let
   version = "7.0.2";
@@ -68,7 +68,7 @@ in stdenv.mkDerivation {
   USE_PYTHON64="n";
   PYTHONLIBTYPE="so";
   PYTHONHOME="${python}";
-  PYTHONVERSION="2.7";
+  PYTHONVERSION=lib.versions.majorMinor python.version;
 
   BLASOPT="-L${blas}/lib -lblas";
   LAPACK_LIB="-L${lapack}/lib -llapack";
--- a/pkgs/apps/pysisyphus/default.nix
+++ b/pkgs/apps/pysisyphus/default.nix
@@ -1,0 +1,140 @@
+{ fetchPypi, fetchFromGitHub, buildPythonPackage, lib, writeTextFile, writeScript, makeWrapper,
+ # Python dependencies
+ autograd, dask, distributed, h5py, jinja2, matplotlib, numpy, natsort, pytest, pyyaml, rmsd, scipy,
+ sympy, scikitlearn, qcengine, ase, xtb-python, openbabel-bindings,
+ # Runtime dependencies
+ runtimeShell, jmol, multiwfn, xtb, openmolcas, pyscf, psi4, wfoverlap, nwchem, orca ? null ,
+ turbomole ? null, gaussian ? null, gamess-us ? null, cfour ? null, molpro ? null,
+ # Test dependencies
+ openssh,
+ # Configuration
+ fullTest ? false
+}:
+let
+  psi4Wrapper = writeScript "psi4.sh" ''
+    #!${runtimeShell}
+    ${psi4}/bin/psi4 -o stdout $1
+  '';
+  pysisrc =
+    let
+      gaussian16Conf = {
+        cmd = "${gaussian}/bin/g16";
+        formchk_cmd = "${gaussian}/bin/formchk";
+        unfchk_cmd = "${gaussian}/bin/unfchk";
+      };
+      text = with lib.lists; lib.generators.toINI {} (builtins.listToAttrs ([
+        { name = "openmolcas"; value.cmd = "${openmolcas}/bin/pymolcas"; }
+        { name = "psi4"; value.cmd = "${psi4Wrapper}"; }
+        { name = "wfoverlap"; value.cmd = "${wfoverlap}/bin/wfoverlap.x"; }
+        { name = "multiwfn"; value.cmd = "${multiwfn}/bin/Multiwfn"; }
+        { name = "jmol"; value.cmd = "${jmol}/bin/jmol"; }
+        { name = "xtb"; value.cmd = "${xtb}/bin/xtb"; }
+      ] ++ optional (gaussian != null) { name = "gaussian16"; value = gaussian16Conf; }
+        ++ optional (orca != null) { name = "orca"; value.cmd = "${orca}/bin/orca"; }
+      ));
+    in
+      writeTextFile {
+        inherit text;
+        name = "pysisrc";
+      };
+
+  binSearchPath = with lib; makeSearchPath "bin" ([
+    jmol
+    multiwfn
+    xtb
+    openmolcas
+    pyscf
+    psi4
+    wfoverlap
+    nwchem
+  ] ++ lists.optional (orca != null) orca
+    ++ lists.optional (turbomole != null) turbomole
+    ++ lists.optional (gaussian != null) gaussian
+    ++ lists.optional (cfour != null) cfour
+    ++ lists.optional (molpro != null) molpro
+  );
+
+in
+  buildPythonPackage rec {
+    pname = "pysisyphus";
+    version = "0.7rc1";
+
+    nativeBuildInputs = [ makeWrapper ];
+
+    propagatedBuildInputs = with lib; [
+      autograd
+      dask
+      distributed
+      h5py
+      jinja2
+      matplotlib
+      numpy
+      natsort
+      pyyaml
+      rmsd
+      scipy
+      sympy
+      scikitlearn
+      qcengine
+      ase
+      xtb-python
+      openbabel-bindings
+      pytest # Also required for normal execution
+      # Syscalls
+      jmol
+      multiwfn
+      xtb
+      openmolcas
+      pyscf
+      psi4
+      wfoverlap
+      nwchem
+    ] ++ lists.optional (orca != null) orca
+      ++ lists.optional (turbomole != null) turbomole
+      ++ lists.optional (gaussian != null) gaussian
+      ++ lists.optional (cfour != null) cfour
+      ++ lists.optional (molpro != null) molpro
+    ;
+
+    src = fetchFromGitHub {
+      owner = "eljost";
+      repo = pname;
+      rev = version;
+      sha256 = "19jwn2hprawz652z0q4d716z2gyb2l3vhjkxn4rjr0dhz5ps4nw2";
+    };
+
+    doCheck = true;
+
+    checkInputs = [ pytest openssh ];
+
+    checkPhase = ''
+      export PYSISRC=${pysisrc}
+      export PATH=$PATH:${binSearchPath}
+      export OMPI_MCA_rmaps_base_oversubscribe=1
+
+      echo $PYSISRC
+
+      ${if fullTest
+          then "pytest -v tests --disable-warnings"
+          else "pytest -v --pyargs pysisyphus.tests --disable-warnings"
+      }
+    '';
+
+    postInstall = ''
+      mkdir -p $out/share/pysisyphus
+      cp ${pysisrc} $out/share/pysisyphus/pysisrc
+      for exe in $out/bin/*; do
+        wrapProgram $exe \
+          --prefix PATH : ${binSearchPath} \
+          --set-default "PYSISRC" "$out/share/pysisyphus/pysisrc"
+      done
+    '';
+
+    meta = with lib; {
+      description = "Python suite for optimization of stationary points on ground- and excited states PES and determination of reaction paths";
+      homepage = "https://github.com/eljost/pysisyphus";
+      license = licenses.gpl3;
+      platforms = platforms.linux;
+      maintainers = [ maintainers.sheepforce ];
+    };
+  }
--- a/pkgs/apps/qcelemental/default.nix
+++ b/pkgs/apps/qcelemental/default.nix
@@ -1,45 +1,41 @@
-{ buildPythonPackage, lib, fetchPypi
-# Other dependencies
-, cacert
-# Python dependencies
-, numpy
-, pydantic
-, pint
-, networkx
-, pytestrunner
-, pytestcov
-, pytest
+{ buildPythonPackage, lib, fetchPypi, fetchpatch, numpy
+, pydantic, pint,  networkx, pytestrunner, pytestcov, pytest
 } :
 
 buildPythonPackage rec {
   pname = "qcelemental";
-  version = "0.19.0";
+  version = "0.20.0";
 
-  checkInputs = [
-    pytestrunner
-    pytestcov
-    pytest
-  ];
-
-  propagatedBuildInputs = [
-    numpy
-    pydantic
-    pint
-    networkx
-    cacert
-  ];
+  checkInputs = [ pytestrunner pytestcov pytest ];
+  propagatedBuildInputs = [ numpy pydantic pint networkx ];
 
   src = fetchPypi {
     inherit pname version;
-    sha256 = "1ljxwhiz1689qijjqzmzydn8q963xx7zxizjnjq0vqg5py9pkyc5";
+    sha256 = "141vw36fmacj897q26kq2bl9l0d23lyqjfry6q46aa9087dcs7ni";
   };
+
+  # FIXME: Fixed upstream but not released yet. Nevertheless critical for correct behaviour.
+  # See https://github.com/MolSSI/QCElemental/pull/265
+  patches = [
+    (fetchpatch {
+      name = "SearchPath1.patch";
+      url = "https://github.com/MolSSI/QCElemental/commit/2211a4e59690bcb14265a60f199a5efe74fe44db.diff";
+      sha256 = "1ibjvmdrc103jdj79xrr11y5yji5hc966rm4ihfhfzgbvfkbjg2l";
+    })
+    (fetchpatch {
+      name = "SearchPath2.patch";
+      url = "https://github.com/MolSSI/QCElemental/commit/5a32ce33e8142047b0a00d0036621fe2750e872a.diff";
+      sha256 = "0gmg70vdps7k6alqclwdlxkli9d8s1fphbdvyl8wy8xrh46jw6rk";
+    })
+  ];
 
   doCheck = true;
 
   meta = with lib; {
-    description = "Periodic table, physical constants, and molecule parsing for quantum chemistry.";
+    description = "Periodic table, physical constants, and molecule parsing for quantum chemistry";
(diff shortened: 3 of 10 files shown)


pysisyphus/default.nix
```diff
@@ -0,0 +1,137 @@
...
+        unfchk_cmd = "${gaussian}/bin/unfchk";
+      };
+      openmolcasConf = {
+        cmd = "${openmolcas}/bin/pymolcas";
+      };
+      orcaConf = {
+        cmd = "${orca}/bin/orca";
+      };
+      psi4Conf = {
+        cmd = "${toString psi4Wrapper}";
```
markuskowa, line 27: This should be sufficient.
```suggestion
        cmd = "${psi4Wrapper}";
```

pysisyphus/default.nix
```diff
@@ -0,0 +1,137 @@
...
+      };
+
+in
+  buildPythonPackage rec {
+    pname = "pysisyphus";
+    version = "0.6";
+
+    nativeBuildInputs = [ makeWrapper ];
+
+    propagatedBuildInputs = with lib.lists; [
```
markuskowa, line 65:
```suggestion
    propagatedBuildInputs = with lib; [
```

markuskowa (changes requested):
pysisyphus itself builds fine for me. However, the test does not build:
```
# (in test file /nix/store/50na6a4b1q59xs0z92f9kwimiw82pk1f-psi4.bats, line 21)
#   `/nix/store/pryg0kagvqgl5bi0p3vrq413k1pjwgk1-python3.8-pysisyphus-0.6/bin/pysis pysis.yaml' failed
# [1619822613.273416] [localhost:49   :0]      tcp_iface.c:602  UCX  ERROR opendir(/sys/class/net) failed: No such file or directory
# /nix/store/nvvrs0sczfwhl45gmwwixzqj7m8x8rpx-python3.8-pyscf-1.7.3/lib/python3.8/site-packages/pyscf/lib/misc.py:47: H5pyDeprecationWarning: Using default_file_mode other than 'r' is deprecated. Pass the mode to h5py.File() instead.
#   h5py.get_config().default_file_mode = 'a'
# Read pysisyphus configuration from '/nix/store/pryg0kagvqgl5bi0p3vrq413k1pjwgk1-python3.8-pysisyphus-0.6/share/pysisyphus/pysisrc'
# xtb-python is not available!
# Traceback (most recent call last):
#   File "/nix/store/pryg0kagvqgl5bi0p3vrq413k1pjwgk1-python3.8-pysisyphus-0.6/bin/..pysis-wrapped-wrapped", line 6, in <module>
#     from pysisyphus.run import run
#   File "/nix/store/pryg0kagvqgl5bi0p3vrq413k1pjwgk1-python3.8-pysisyphus-0.6/lib/python3.8/site-packages/pysisyphus/run.py", line 25, in <module>
#     from pysisyphus.calculators import *
#   File "/nix/store/pryg0kagvqgl5bi0p3vrq413k1pjwgk1-python3.8-pysisyphus-0.6/lib/python3.8/site-packages/pysisyphus/calculators/__init__.py", line 38, in <module>
#     from pysisyphus.calculators.ONIOMv2 import ONIOM
#   File "/nix/store/pryg0kagvqgl5bi0p3vrq413k1pjwgk1-python3.8-pysisyphus-0.6/lib/python3.8/site-packages/pysisyphus/calculators/ONIOMv2.py", line 48, in <module>
#     from pysisyphus.calculators.PySCF import PySCF
#   File "/nix/store/pryg0kagvqgl5bi0p3vrq413k1pjwgk1-python3.8-pysisyphus-0.6/lib/python3.8/site-packages/pysisyphus/calculators/PySCF.py", line 7, in <module>
#     from pyscf.dft import xcfun
#   File "/nix/store/nvvrs0sczfwhl45gmwwixzqj7m8x8rpx-python3.8-pyscf-1.7.3/lib/python3.8/site-packages/pyscf/dft/xcfun.py", line 32, in <module>
#     _itrf = lib.load_library('libxcfun_itrf')
#   File "/nix/store/nvvrs0sczfwhl45gmwwixzqj7m8x8rpx-python3.8-pyscf-1.7.3/lib/python3.8/site-packages/pyscf/lib/misc.py", line 69, in load_library
#     return numpy.ctypeslib.load_library(libname, _loaderpath)
#   File "/nix/store/cx0py33vl5li9yinz67rbky8g0lrv3wf-python3.8-numpy-1.20.2/lib/python3.8/site-packages/numpy/ctypeslib.py", line 153, in load_library
#     raise OSError("no file with expected extension")
# OSError: no file with expected extension
builder for '/nix/store/rpvm5fnivwvpmf25jmn2y4zyc5xzmmdk-psi4.drv' failed with exit code 1
```

sheepforce:
At the moment I cannot reproduce the error, as for me the PySCF dependency already fails. I've tried with disabled PySCF tests and still get the AM1 errors and H5Py deprecation warnings when trying to run Pysisyphus. I will have a look as soon as PySCF is working.

sheepforce:
So this now makes pysisyphus working :slightly_smiling_face: A minor PySCF update was necessary due to [Pyscf/#872](https://github.com/pyscf/pyscf/issues/872). I've also added the XTB-python library for more efficient interfacing.

The tests from the check phase do not work within nix right now and I guess this is due to how they are written.

@markuskowa I've added bats test. The Molcas and Turbomole ones do not work in the bats test, while they work fine in a nix-shell. The actually throw funny python errors. Do you have an idea what causes trouble here?

@eljost Most of the assertions do not pass. XTB needs more cycles and converges to a higher energy, The Molcas does not converge and does not look as if it ever would and the Turbomole test raises a `KeyError 'xyz'`. Could you please have a look what's going on there?

markuskowa:
Under which conditions is the molcas test supposed to pass?
The molcas test also fails in Hydra:
```
#                          -----------------------------
#                          | HESSIAN AT FINAL GEOMETRY |
#                          -----------------------------
#
# ... mass-weighing cartesian hessian
# ... doing Eckart-projection
#
# First 10 eigenvalues [-1.4938e-02 -2.2596e-17 -4.0637e-18 -1.2579e-18  5.9134e-19  1.3083e-17
#   2.1119e-17  1.0916e-05  2.4721e-05  2.6807e-05]
# Imaginary frequencies: [-627.83] cm⁻¹
#
# Wrote final (not mass-weighted) hessian to 'calculated_final_cart_hessian'.
# Wrote HD5 Hessian to 'final_hessian.h5'.
# Wrote imaginary mode with ṽ=-627.83 cm⁻¹ to 'imaginary_mode_000.trj'
#
# Barrier between TS and first COS image: 55.8 kJ mol⁻¹
# Barrier between TS and last COS image: 138.1 kJ mol⁻¹
#
#
#                              #####################
#                              # ASSERTING RESULTS #
#                              #####################
# 00: opt.cur_cycle
#       Reference: 24
#         Current: 24
#         Matches: True
# 01: ts_opt.is_converged
#       Reference: True
#         Current: True
#         Matches: True
# 02: ts_opt.cur_cycle
#       Reference: 10
#         Current: 15
#         Matches: False
# 03: ts_geom._energy
#       Reference: -80.70048732
#         Current: -80.680988069325
#         Matches: False
# Traceback (most recent call last):
#   File "/nix/store/01cljm6c2jshiksslk3966m7sgf2nyzs-python3.8-pysisyphus-0.6/bin/..pysis-wrapped-wrapped", line 9, in <module>
#     sys.exit(run())
#   File "/nix/store/01cljm6c2jshiksslk3966m7sgf2nyzs-python3.8-pysisyphus-0.6/lib/python3.8/site-packages/pysisyphus/run.py", line 1946, in run
#     return run_from_dict(run_dict, **run_kwargs)
#   File "/nix/store/01cljm6c2jshiksslk3966m7sgf2nyzs-python3.8-pysisyphus-0.6/lib/python3.8/site-packages/pysisyphus/run.py", line 1902, in run_from_dict
#     check_asserts(run_result, run_dict)
#   File "/nix/store/01cljm6c2jshiksslk3966m7sgf2nyzs-python3.8-pysisyphus-0.6/lib/python3.8/site-packages/pysisyphus/run.py", line 1633, in check_asserts
#     assert all(matches)
# AssertionError
builder for '/nix/store/pxl50w4kk83prk0035vkag46imp4ssa6-psi4.drv' failed with exit code 1
```

eljost:
> @eljost Most of the assertions do not pass. XTB needs more cycles and converges to a higher energy, The Molcas does not converge and does not look as if it ever would and the Turbomole test raises a `KeyError 'xyz'`. Could you please have a look what's going on there?

Which tests do not pass? The files under `examples` are not part of my CI test suite, so maybe not all of the examples there are up to date...

sheepforce:
@eljost Here is a quick list of things from the examples that fail with outputs:
- [01_turbomole_cytosin_s1_opt](https://github.com/markuskowa/NixOS-QChem/files/6521199/01_turbomole_cytosin_s1_opt.out.txt) `KeyError: 'xyz'`
- [03_xtb_neb_optimization](https://github.com/markuskowa/NixOS-QChem/files/6521242/03_xtb_neb_optimization.out.txt) more cycles, higher energy and wild XTB timings
- [07_openmolcas_trans_butadiene_s1_opt](https://github.com/markuskowa/NixOS-QChem/files/6521315/07_openmolcas_trans_butadiene_s1_opt.out.txt) [optimizer.log](https://github.com/markuskowa/NixOS-QChem/files/6521323/optimizer.log) (incomplete) doesn't look as if it would go anywhere in the optimisation. Is quite far of still in the energies from what the assertion expects

@markuskowa I guess the tests are always supposed to pass, but let's wait for what @eljost can tell us there. Regarding the Molcas example (you got the XTB example); it runs in a `nix-shell` with the output above, but it produces a completely different error in the bats test: [molcas_bats.txt](https://github.com/markuskowa/NixOS-QChem/files/6521337/molcas_bats.txt)

markuskowa:
> but it produces a completely different error in the bats test: molcas_bats.txt.

The snippet above is taken from the hydra build of `qchem.tests.pysisyphus`.

> I guess the UCX error right at the start could be the problem? Although I have no idea why this occurs.

This is a sandbox issue, since there is no `/sys` mounted. I have seen this in other builds, but I do not think it is a problem here.

sheepforce:
> The snippet above is taken from the hydra build of `qchem.tests.pysisyphus`.

Ah yes. It's the XTB part that fails the assertion error in your previous post. This has been fixed now. But also the Molcas test failed with an error that only happens in the bats test.

However, I sat down with @eljost and we had a look at the test suite. The assumption that they are only executed from within a Git repo has been fixed. The pytest now actually runs properly with Nix. At the moment we are pinning a commit from a hotfix branch, but before merging we can wait for a new release from @eljost with the fixes included.

@eljost The full test suite still has some failures. Can you have a look? You can execute/build/test like this `nix-build`
[`pysisTest.nix`](https://github.com/markuskowa/NixOS-QChem/files/6523399/pysisTest.nix.txt)`

`

pkgs/apps/nwchem/default.nix
```diff
@@ -3,7 +3,7 @@
 , automake, autoconf, libtool, makeWrapper
 } :
 
-assert blas.isILP64 == blas.isILP64;
```
markuskowa, line 6: true ;)

pkgs/apps/pysisyphus/default.nix
```diff
@@ -0,0 +1,158 @@
...
+        cmd = "${wfoverlap}/bin/wfoverlap.x";
+      };
+      multiwfnConf = {
+        cmd = "${multiwfn}/bin/Multiwfn";
+      };
+      jmolConf = {
+        cmd = "${jmol}/bin/jmol";
+      };
+      text = with lib.lists; lib.generators.toINI {} (builtins.listToAttrs ([
+        { name = "openmolcas"; value = openmolcasConf; }
```
markuskowa, line 47:
```suggestion
        { name = "openmolcas"; value.cmd = "${openmolcas}/bin/pymolcas"; }
```
This pattern would save a lot of space here.

pkgs/apps/xtb/default.nix
```diff
@@ -1,18 +1,30 @@
-{ stdenv, lib, gfortran, fetchFromGitHub, cmake, makeWrapper, mkl
-,  turbomole ? null, orca ? null, cefine ? null
-} :
+{ stdenv, lib, gfortran, fetchFromGitHub, cmake, makeWrapper, writeTextFile, mkl,
```
markuskowa, line 1:
```suggestion
{ stdenv, lib, gfortran, fetchFromGitHub, cmake, makeWrapper, writeTextFile, mkl
, ...
} :
```
nixpkgs style

pkgs/apps/xtb/default.nix
```diff
@@ -23,25 +35,38 @@ stdenv.mkDerivation rec {
...
-    # programs that XTB might call.
-    binSearchPath = with lib; strings.makeSearchPath "bin" ([ ]
-      ++ lists.optional (turbomole != null) turbomole
-      ++ lists.optional (orca != null) orca
-      ++ lists.optional (turbomole != null) cefine
-    );
-  in  ''
+  postInstall = ''
+    mkdir -p $out/lib/pkgconfig
+    substitute ${xtbPkgConfig} $out/lib/pkgconfig/xtb.pc \
```
markuskowa, line 51: Instead of creating a file in the nix-store first, I would write the file directly in the `postInstall` phase:
```
cat > $out/lib/pkgconfig/xtb.pc << EOF
...
EOF
```
Here you have shell substitution of `$out` automatically.

pkgs/apps/xtb/default.nix
```diff
@@ -23,25 +35,38 @@ stdenv.mkDerivation rec {
...
+      --subst-var out
+  '';
+
+  # Syscalls that XTB might make.
+  binSearchPath = with lib; strings.makeSearchPath "bin" ([ ]
```
markuskowa, line 56: `binSearchPath` does not need to available in the derivation?
better:
```
postFixup = let 
  binSearchPath = ...;
in ''
''
wrapProgram ....
'';

File: pkgs/lib/rmsd/default.nix [hunk 8]
```diff
@@ -0,0 +1,12 @@
...
+buildPythonPackage rec {
+    pname = "rmsd";
+    version = "1.4";
+
+    propagatedBuildInputs = [ scipy ];
+
+    src = fetchPypi  {
+      inherit pname version;
+      sha256 = "0gsncsszya1zp2yw777f46df2a3wk80d8np5c4ihzalzj5glma0n";
+    };
```
markuskowa, line 11: `meta` = ?

pythonPackages.nix
```diff
@@ -47,6 +47,18 @@ let
       rev = "9b60184c5d161e4871c91ce29a44e3ac2c2a438e";
       sha256 = "1vh8dp3nw4fk1mnfv0w8ici1lzxyfn5han7hipqzsfxl75w76r18";
     };
+
+    pysisyphus = callPackage ./pkgs/apps/pysisyphus {
+      orca = selfPkgs.orca;
+      turbomole = selfPkgs.turbomole;
+      cfour = selfPkgs.cfour;
+      # gaussian = if cfg.optpath != null then self.gaussian else null;
+      # molpro = if cfg.licMolpro != null then self.molpro else null;
```
markuskowa, line 56: `molpro` is now `null` if no license key is provided. We can drop the conditionals checking for `cfg.licMolpro`.

sheepforce:
@markuskowa All good suggestions :slightly_smiling_face: I will take care of those when Pysis actually works. More and more strange behaviour is popping up here. The Turbomole tests only pass if CFour is present and stuff like this ... 

I will probably meet with @eljost next week and we will try to get to solve the remaining issues and then I will finalise the PR.

pythonPackages.nix
sheepforce: fixed for molpro but `gaussian` is still an unknown attribute to `selfPkgs`. This is simply missing if `optpath` is not set, right?

pkgs/apps/pysisyphus/default.nix
```diff
@@ -0,0 +1,138 @@
...
+ bash, jmol, multiwfn, xtb, openmolcas, pyscf, psi4, wfoverlap, nwchem, orca ? null ,
+ turbomole ? null, gaussian ? null, gamess-us ? null, cfour ? null, molpro ? null,
+ # Test dependencies
+ openssh,
+ # Configuration
+ fullTest ? false
+}:
+let
+  psi4Wrapper = writeScript "psi4.sh" ''
+    #!${bash}/bin/bash
```
markuskowa, line 15:
```suggestion
    #!${runtimeShell}
```

pkgs/apps/pysisyphus/default.nix
```diff
@@ -0,0 +1,138 @@
...
+      export PATH=$PATH:${binSearchPath}
+      export OMPI_MCA_rmaps_base_oversubscribe=1
+
+      ${if fullTest
+          then "pytest -v tests --disable-warnings"
+          else "pytest -v --pyargs pysisyphus.tests --disable-warnings"
+      }
+    '';
+
+    binSearchPath = with lib; makeSearchPath "bin" ([
```
markuskowa, line 105: That is better placed in the top level `let`.

pkgs/apps/qcelemental/default.nix
```diff
@@ -1,45 +1,30 @@
...
 
   src = fetchPypi {
     inherit pname version;
-    sha256 = "1ljxwhiz1689qijjqzmzydn8q963xx7zxizjnjq0vqg5py9pkyc5";
+    sha256 = "141vw36fmacj897q26kq2bl9l0d23lyqjfry6q46aa9087dcs7ni";
   };
 
+  # FIXME: Fixed upstream but not released yet. Nevertheless critical for correct behaviour.
+  # See https://github.com/MolSSI/QCElemental/pull/265
+  patches = [ ./SearchPath.patch ];
```
markuskowa, line 19: When the upstream fix is merged we can remove `qcelemental` and the patches.

sheepforce: It is not in nixpkgs-unstable yet so I've kept it for the moment in the same version as on nixpkgs master?

pythonPackages.nix
sheepforce: I've did the same to gaussian as you did to molpro as a workaround :slightly_smiling_face:

sheepforce:
OK, thanks to @eljost we could figure out a lot of problems, including the QCElemental bug and now it's running just fine. From my perspective we are good to go. :slightly_smiling_face:

pkgs/apps/qcelemental/default.nix
markuskowa: Yes, that is the disadvantage of `nixpkgs-unstable` here. It can take a few days while until the channels advances.

markuskowa (approved):

Issue #24, opened 2021-04-29
Explicit MPI Binaries


sheepforce:
There is the awkward problem for software, that directly relies on calling the MPI executables (such as BAGEL, NWChem, CP2K, ...). Linking the mpi executables to the `bin` directory of the derivation potentially leads to multiple executables with the same name and therefore conflicts, while not providing them with the derivation requires the user to add the correct MPI version to the nix-shell (or whatever), which is especially difficult if an override to the MPI derivation has been used.

One option is to wrap the programs, so that the wrapper calls MPI on the actual executable, [as you did with NWChem](https://github.com/markuskowa/NixOS-QChem/blob/23e8a29b5e23e2c508fae1582a10b64c62a7d0f6/nwchem/default.nix#L115).
This is very easy to use and always works but has the disadvantage that it is not possible to pass additional parametres to MPI (such as `--mca` to select the network, or `-N` to select the distribution of processes between nodes, which I often require for multi-node jobs).

An alternative that would only work with `mkShell` could be to use `passthru` in the derivation to provide a shell-hook, like this:
```nix
{ stdenv, mpi, ... }:

stdenv.mkDerivation rec {
  pname = "nwchem";
  ...
  passthru = {
    mpiHook = ''
      export ${pname}_MPIFLAVOUR=${mpi.pname}
      export ${pname}_MPIEXEC=${mpi}/bin/mpiexec
    '';
  };
}
```
and then, in a `nix-shell` use it somehow like this:
```nix
mkShell {
  buildInputs = [ nwchem ];
  shellHook = ''
    ${nwchem.passthru.mpiHook}
  '';
}
```
with `nix-shell --run "$nwchem_MPIEXEC -n 4 -N 4 --mca btl openib,self nwchem input > output"`?

The disadvantage here is, that I don't see how one could use this without the `shellHook` of a `mkShell`.

markuskowa:
Yes, that only works with `mkShell`. A `setup-hook` would be more universal here but is also not a pretty solution. It means more code. Every MPI program needs to define it in a uniform way. An easier solution would be to just create a symlink pointing to mpiexec/mpirun: `mpirun.nwchem -> /nix/store/mpi/bin/mpirun` - workable but also not very clean. 
The wrapper scripts I used for NWchem were a quick'n dirty solution that will disappear in the future.

[...]

PR #32, opened 2021-04-30
MPI passthru

sheepforce:
Adds the MPI passthru as discussed in #24 as a first step towards a solution of MPI calls.

### DIFF
--- a/bagel/default.nix
+++ b/bagel/default.nix
@@ -96,6 +96,8 @@ in stdenv.mkDerivation {
 
   doCheck = true;
 
+  passthru = { inherit mpi; };
+
   meta = with lib; {
     description = "Brilliantly Advanced General Electronic-structure Library";
     homepage = "https://nubakery.org";
@@ -104,4 +106,3 @@ in stdenv.mkDerivation {
     platforms = [ "x86_64-linux" ];
   };
 }
-
--- a/default.nix
+++ b/default.nix
@@ -70,10 +70,12 @@ let
 
         configureFlags = with lib.lists; oldAttrs.configureFlags ++ [
           "--enable-mpi"
-          "MPICC=${self_.mpi}/bin/mpicc"
-          "MPIFC=${self_.mpi}/bin/mpif90"
-          "MPIF90=${self_.mpi}/bin/mpif90"
+          "MPICC=${self.mpi}/bin/mpicc"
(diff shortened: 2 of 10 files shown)


bagel/default.nix
```diff
@@ -96,6 +96,8 @@ in stdenv.mkDerivation {
 
   doCheck = true;
 
+  passthru.mpi = { inherit mpi; };
```
markuskowa, line 99: That results in `passthru.mpi.mpi` but we want `passthru.mpi`. It should read `passthru.mpi = mpi` ( or  `passthru = {inherit mpi};` if you want to use the `inherit` )

default.nix
```diff
@@ -70,10 +70,12 @@ let
...
           "--enable-mpi"
-          "MPICC=${self_.mpi}/bin/mpicc"
-          "MPIFC=${self_.mpi}/bin/mpif90"
-          "MPIF90=${self_.mpi}/bin/mpif90"
+          "MPICC=${self.mpi}/bin/mpicc"
+          "MPIFC=${self.mpi}/bin/mpif90"
+          "MPIF90=${self.mpi}/bin/mpif90"
         ];
+
+        passthru.mpi = { mpi = self.mpi; };
```
markuskowa, line 78: For a library it is not really needed. Putting `mpi` in `propagatedBuildInputs` is probably more useful in this case.

nixpkgs-opt.nix
```diff
@@ -27,6 +27,8 @@ self: super:
       "MPIFC=${self.mpi}/bin/mpif90"
       "MPIF90=${self.mpi}/bin/mpif90"
     ];
+
+    propagatedBuildInputs = oldAttrs.propagatedBuildInputs ++ [self.mpi]
```
markuskowa, line 31:
```suggestion
    propagatedBuildInputs = oldAttrs.propagatedBuildInputs ++ [ self.mpi ];
```

gdma/default.nix
```diff
@@ -32,6 +32,8 @@ stdenv.mkDerivation rec {
 
     hardeningDisable = [ "format" ];
 
+    passthru = { inherit mpi; };
```
markuskowa, line 35: `gdma` does not use MPI.

sheepforce: Sorry, this one slipped in. Must have confused it with GPAW.

Issue #33, opened 2021-05-03
CP2K OMP Instability/Wrong


sheepforce:
I started using CP2K from the overlay on a simple system and noticed that it does not execute properly when using hybrid parallelisation. Whenever the number of OMP threads /= 1, cholesky decomposition will fail and it also becomes very slow. This usually happens when some threading problems with underlying BLAS occur. I've tested with the defaults but also with MKL as BLAS and LAPACK provider, but the results are the same. Both one of my "real life" examples as well as many inputs from the test suite fail when using OMP parallelisation.

`OMP_NUM_THREADS=4 cp2k` [`pbe_nvt.txt`](https://github.com/markuskowa/NixOS-QChem/files/6416528/pbe_nvt.txt).

Pure MPI execution works fine. Nevertheless, for the psmp version, that is being built, the combination of both should work fine (and actually does on the same version of CP2K built without Nix).

I couldn't narrrow it down further, yet.

sheepforce:
I was able to narrow it down further. It actually is the BLAS threading. For reasons unknown to me, the abstract `blas` and `lapack` derivations both do not work. Unwrapped `openblasCompat`, `amd-blis` and `amd-libflame` also do not work. The only thing where I could get correct threading behaviour is with unwrapped `mkl`. It is not enough to simply do `blas.override { blasProvider = super.mkl; }` and `lapack.override { lapackProvider = super.mkl; }`, it really needs to be provided unwrapped.

I will make a PR with more details, but it is not too nice, as it introduces many conditionals.

markuskowa:
What advantage does the OpenMP parallelization in CP2K bring? I did benchmarks against NixOS 20.03, and I couldn't see any gains (with the test cases in NixOS-QChem). Maybe should just drop the OpenMP support and build it purely with MPI?

sheepforce:
Here the hybrid parallelisation is actually very helpful, especially for large calculations with Hartree-Fock exchange or RI-MP2. It saves memory and also communication overhead if the calculations become very large. I have good experience with the hybrid scheme for calculations on our cluster, that are parallel over >= 10 nodes (360 cores used with 4 OMP  threads, 90 MPI processes). This was nearly saving a factor of 2 in time.

sheepforce:
I just realised that mpi in nixpkgs is built without threading support 🤔 This might be the cause for the problems here. `--enable-mpi-thread-multiple` is usually required for openmpi to support the hybrid scheme and does no harm otherwise.

[...]

PR #34, opened 2021-05-04
cp2k: fix openmp threading

sheepforce:
This works around threading problems in CP2K for the PSMP version (see #33). Depending on the BLAS and LAPACK providers either the `psmp` version (for MKL only) or the `popt` version (for anything else) is built.

In principle it also should work with OpenBLAS, at least I've used CP2K 7.1 with OpenBLAS in the PSMP version on our cluster.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -118,7 +118,7 @@ let
 
       chemps2 = callPackage ./chemps2 {};
 
-      cp2k = callPackage ./cp2k rec {
+      cp2k = callPackage ./cp2k {
         libxc = self.libxc4;  # patches are are required for libxc5
         inherit optAVX;
       };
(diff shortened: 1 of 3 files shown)


sheepforce:
The only version that always gives correct results is the one with MKL. I couldn't find any other option, that always gives correct results. Therefore I would suggest to always build CP2K with MKL for the moment.

default.nix
```diff
@@ -118,7 +118,7 @@ let
 
       chemps2 = callPackage ./chemps2 {};
 
-      cp2k = callPackage ./cp2k {
+      cp2k = callPackage ./cp2k rec {
```
markuskowa, line 121:
```suggestion
      cp2k = callPackage ./cp2k {
```

PR #35, opened 2021-05-05
mvapich: allow for different network types

sheepforce:
Adds options to MVAPICH to be used with different network types (ethernet, Infiniband and Intel's OmniPath). The test suite of MVAPICH passes but my two test cases (scalapack and Dalton) do not work properly with it. Both segfault (scalapack only for a few tests, not for all). I am not certain why. The [derivation that I've used in my overlay](https://gitlab.com/theoretical-chemistry-jena/nixwithchemistry/-/blob/master/pkgs/libs/mvapich2/default.nix) is very similar and didn't segfault, but I also ever used only nixos-20.09 channel. I am not entirely sure if this is again a threading problem with BLAS. @markuskowa Do you have any suggestion for a package, that uses MPI but not BLAS and LAPACK, where I could narrow this down? It is at least not one of the `configureFlags` of the derivation. I've also tried with only `--with-device=ch3:sock` and nothing else to be able to run anything on my workstation , but this still segfaults.

Update: also segfaults for nixos-20.09

### DIFF
--- a/mvapich/default.nix
+++ b/mvapich/default.nix
@@ -12,18 +12,6 @@
 
 assert builtins.elem network [ "ethernet" "infiniband" "omnipath" ];
 
-assert
-  (if network == "infiniband"
-     then (lib.lists.all (x: x != null) [ infiniband-diags opensm rdma-core ])
-     else true
-  );
-
-assert
-  (if network == "omnipath"
-     then (lib.lists.all (x: x != null) [ libpsm2 libfabric ])
-     else true
-  );
-
 let
 
 
@@ -44,7 +32,7 @@ in stdenv.mkDerivation rec {
 
   nativeBuildInputs = [ pkgconfig bison makeWrapper ];
   propagatedBuildInputs = [ numactl rdma-core zlib infiniband-diags opensm ];
-  buildInputs = [
+  buildInputs = with lib; [
     numactl
     libxml2
     perl
@@ -52,12 +40,9 @@ in stdenv.mkDerivation rec {
     slurm
     openssh
     hwloc
-    rdma-core
-    infiniband-diags
-    opensm
-    libpsm2
-    libfabric
-  ];
+  ] ++ lists.optionals (network == "infiniband") [ rdma-core infiniband-diags opensm ]
+    ++ lists.optionals (network == "omnipath") [ libpsm2 libfabric ]
+  ;
 
   configureFlags = [
     "--with-pm=hydra"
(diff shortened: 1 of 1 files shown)


mvapich/default.nix
```diff
@@ -1,10 +1,29 @@
...
 , useSlurm ? false
+# Network type for MVAPICH2
+, network ? "ethernet"
 } :
 
+assert builtins.elem network [ "ethernet" "infiniband" "omnipath" ];
+
+assert
+  (if network == "infiniband"
+     then (lib.lists.all (x: x != null) [ infiniband-diags opensm rdma-core ])
```
markuskowa, line 17: I don't think this is needed. We always hand over these input parameters and non-null values. This assert and the next one can be removed.

mvapich/default.nix
```diff
@@ -23,15 +42,35 @@ in stdenv.mkDerivation rec {
...
 
-  configureFlags = lib.optionals useSlurm [ "--with-pmi=pmi1" "--with-pm=slurm" ];
+  configureFlags = [
+    "--with-pm=hydra"
+    "--enable-fortran=all"
+    "--enable-cxx"
+    "--enable-threads=multiple"
+    "--enable-hybrid"
+    "--enable-shared"
+  ] ++ lib.optionals useSlurm [ "--with-pmi=pmi1" "--with-pm=slurm" ]
```
markuskowa, line 69: While we're at it: this should be pmi2 (or pmix if supported)

mvapich/default.nix
```diff
@@ -23,15 +42,35 @@ in stdenv.mkDerivation rec {
...
-                  infiniband-diags opensm ];
+  buildInputs = [
+    numactl
+    libxml2
+    perl
+    gfortran
+    slurm
+    openssh
+    hwloc
+    rdma-core
```
markuskowa, line 55: It would make sense to also make the packages for respective interfaces conditional depending on which interface type is chosen. That may avoid that configure picks up unnecessary packages.

markuskowa:
One difference to you derivation, that I can see right now, is that it does not use `hwloc`.
I don't have good example for an app that uses MPI but no BLAS/LAPACK. However, you could try HPL, which is rather simple and only uses basic BLAS routines.

sheepforce:
I've did a little bit more testing. I've built MVAPICH and Scalapack in a KVM on Debian Buster and on our CentOS cluster with basically the same flags that nix is passing to `configure` and `cmake` (just changed paths). 

- GCC/GFortran 8.3.0 (self built on CentOS, from repos on Debian)
- OpenBLAS (0.3.6 on CentOS, 0.3.5 on debian) with LP64 interface and `USE_OPENMP=1` in both cases.
- GLibC 2.17 on CentOS, GLibC 2.28 on Debian
- Kernel 3.10 on CentOS, 4.19 on Debian

On Debian I get exactly the same segfaults as with Nix, on CentOS everythings works fine. So now things become a little bit curious. There is a huge version jump in GLibC and the Kernel, while the compilers and BLAS, that I suspected are very similar (and I don't see too many other influences). A this point I suspect, that this is a problem upstream could have.

markuskowa:
It could be an issue with newer glibc (this would be my guess here). The kernel is less likely to cause such issues. 
Another difference between your working derivation and this one is the version (2.3.4 <-> 2.3.5). Could this be it?

Otherwise am not sure how to fix this right now. We can merge your changes anyway since it is an improvement for the derivation. I would then simply wait for next release of MVAPICH and see what happens.

sheepforce:
Alright, I will open the PR for merging. I would be surprised if it would be 2.3.4 <-> 2.3.5 but I will check.

PR #42, opened 2021-05-12
release.nix: fixing proprietary package access

sheepforce:
Makes also the mandatory tests optional, if the proprietary packages are not available.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -53,11 +53,10 @@ let
             tests.molcas
           ] ++ pkgSet.lib.optionals allowUnfree [
             molden
-          ] ++ pkgSet.lib.optionals (cfg.srcurl != null && allowUnfree) [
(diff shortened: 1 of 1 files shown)


markuskowa:
I am working on a solution which filters out every package that is marked as `unfree` in `release.nix`. That way we can get rid of all the `allowUnfree` conditionals in the overlay code. However, it is not working as intended yet.

sheepforce:
Alright, then I will stand by :slightly_smiling_face: I was actually not looking for an option to completely disable unfree packages here (which is also a good option to have), but just a mechanism to disable some packages, which I don't have access to, while keeping other unfree ones, that I still wanted to test. So I wanted to keep Turbomole and CFour tests but disable Molpro and others.

PR #43, opened 2021-05-12
Openmolcas Tag Fix & MKL Support

sheepforce:
Now that I have a running Hydra I noticed that OpenMolcas 20.10 does not build. The Molcas version hashes are not stable. `v20.10` was moved to a different commit multiple times (which one does not realise if the sources are cached in the store). I fixed this by pinning to the current commit hash of the tags, as the tags of molcas are not really stable. For the current molcas 21.02 i also enabled building with MKL.

### DIFF
--- a/pkgs/apps/openmolcas/v20.10.nix
+++ b/pkgs/apps/openmolcas/v20.10.nix
@@ -4,7 +4,7 @@
 } :
 
 let
-  version = "20.10";
+  version = "20.10-14.10.2020";
   gitLabRev = "v${version}";
 
   python = python3.withPackages (ps : with ps; [ six pyparsing ]);
@@ -32,6 +32,7 @@ in stdenv.mkDerivation {
     url = "https://raw.githubusercontent.com/NixOS/nixpkgs/2eee4e4eac851a2846515dcfa3274c4ab92ecbe5/pkgs/applications/science/chemistry/openmolcas/openblasPath.patch";
     sha256 = "0l6z5zhfbfpbp9x58228nhhwwp1fzmi8cmmasvzddp84h31f0b8h";
   })
+    ./MKL-MPICH.patch
   ];
 
   prePatch = ''
--- a/pkgs/apps/openmolcas/v20.10.nix
+++ b/pkgs/apps/openmolcas/v20.10.nix
@@ -23,8 +23,8 @@ in stdenv.mkDerivation {
   src = fetchFromGitLab {
     owner = "Molcas";
     repo = "OpenMolcas";
-    rev = gitLabRev;
-    sha256 = "1w8av44dx5r9yp2xhf9ypdrhappvk984wrd5pa1ww0qv6j2446ic";
+    rev = "01fe7a2bdca51c51d183f3061b0ee9c631bf9bec";
+    sha256 = "0xr9plgb0cfmxxqmd3wrhvl0hv2jqqfqzxwzs1jysq2m9cxl314v";
   };
 
   patches = [ (fetchpatch {
@@ -126,4 +126,3 @@ in stdenv.mkDerivation {
     platforms = platforms.linux;
   };
 }
-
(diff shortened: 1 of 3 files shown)


pkgs/apps/openmolcas/v21.02.nix
```diff
@@ -40,9 +49,22 @@ in stdenv.mkDerivation {
     chmod -R u+w External/
   '';
 
+  postPatch = ''
+    substituteInPlace CMakeLists.txt \
+      --replace 'set (libpath "''${MKLROOT}/lib/intel64")' 'set (libpath "''${MKLROOT}/lib")' \
+      --replace 'find_library (LIBMKL_BLACS NAMES "mkl_blacs_ilp64"' 'find_library (LIBMKL_BLACS NAMES "mkl_blacs_intelmpi_ilp64"'
```
markuskowa, line 55: This requires MPICH/MVAPICH/IntelMPI? There is also a `mkl_blacs_openmpi_ilp64`. Does this work properly?

pkgs/apps/openmolcas/v20.10.nix
```diff
@@ -23,8 +23,8 @@ in stdenv.mkDerivation {
   src = fetchFromGitLab {
     owner = "Molcas";
     repo = "OpenMolcas";
-    rev = gitLabRev;
-    sha256 = "1w8av44dx5r9yp2xhf9ypdrhappvk984wrd5pa1ww0qv6j2446ic";
+    rev = "01fe7a2bdca51c51d183f3061b0ee9c631bf9bec";
+    sha256 = "0xr9plgb0cfmxxqmd3wrhvl0hv2jqqfqzxwzs1jysq2m9cxl314v";
```
markuskowa, line 27: Quiet hash changes are a nightmare for packaging...

pkgs/apps/openmolcas/v21.02.nix
```diff
@@ -1,11 +1,20 @@
...
+  (blas-i8.isILP64 || blas-i8.passthru.implementation == "mkl")
+  "A 64 bint integer BLAS implementation is required.";
+
+assert
+  lib.asserts.assertMsg
+  (builtins.elem blas-i8.passthru.implementation [ "openblas" "mkl" ])
+  "OpenMolcas requires OpenBLAS or MKL.";
+
 let
   version = "21.02";
```
markuskowa, line 17: I wonder if we should also append a date string to the version (`.YYYY-MM-DD`) to bring more clarity into the inconsistent versioning?

pkgs/apps/openmolcas/v21.02.nix
sheepforce: Luckily it does not :slightly_smiling_face: Those are [these lines here](https://gitlab.com/Molcas/OpenMolcas/-/blob/master/CMakeLists.txt#L1238). OpenMPI should already work correctly without the second `--replace` line (If we decide to add MPI support to Molcas later (we can port it from my overlay))

pkgs/apps/openmolcas/v20.10.nix
sheepforce: I know ... I ran into this multiple times with my own molcas versions :disappointed:

pkgs/apps/openmolcas/v21.02.nix
sheepforce: Might be a good idea. I guess they are using their moving tags just for minor patch updates (at least from what I've observed so far). I guess nothing should change results there, but at least it changed it from "does not build with this feature"  to "builds with this feature" a few times ...

pkgs/apps/openmolcas/v21.02.nix
markuskowa: Since the substitution here is static it would be better to provide it as patch. `substituteInPlace` does not fail when the change becomes obsolete but a patch does. Then we know that it can be removed (or needs to be adapted).

pkgs/apps/openmolcas/v21.02.nix
markuskowa: Then let's add the date strings to the version. This brings in more clarity. Ideally the molcas developer would create a proper release with patchlevels (i.e. 21.02.X) but since this is not the case we need to work around it.

sheepforce:
Done

PR #44, opened 2021-05-14
release: provide a cleaner way to remove unfree/private source packages

markuskowa:
* In case `allowUnfree` is set to `false`, packages that are unfree or depend on unfree packages are now removed from the package set without creating eval errors. This will allow for a clean Hydra jobset which builds only the opensource source software.
* The `tested` constituents are now listed in form of strings and picked from the package set. In case a package has been removed from the package set (i.e. by setting it to `null`) it will be ignored without an eval error.

Closes https://github.com/markuskowa/NixOS-QChem/pull/42

@sheepforce could you test if this evaluates now on your setup?

### DIFF
--- a/pkgs/apps/qdng/default.nix
+++ b/pkgs/apps/qdng/default.nix
@@ -27,10 +27,11 @@ in stdenv.mkDerivation {
                   bzip2 zlib libxml2 flex bison ];
   nativeBuildInputs = [ automake autoconf libtool ];
 
-  meta = {
+  meta = with lib; {
     description = "Quantum dynamics program package";
-    platforms = lib.platforms.linux;
-    maintainer = "markus.kowalewski@gmail.com";
+    platforms = platforms.linux;
+    maintainer = [ maintainers.markuskowa ];
+    license = licenses.unfree;
   };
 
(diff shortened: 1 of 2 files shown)


sheepforce:
The `null` settings for Molpro and others work now, thanks :slightly_smiling_face: What is not working is setting `scrurl` to a file:
```
trying file:///mnt/sol/turbolinux751.tar.gz
curl: (37) Couldn't open file /mnt/sol/turbolinux751.tar.gz
error: cannot download turbolinux751.tar.gz from any mirror
builder for '/nix/store/6jdacpy3va3kvik136ijjkhzbpwsipiz-turbolinux751.tar.gz.drv' failed with exit code 1
```
and similar for the other packages with `requireFile`. With `config = { optAVX=true; srcurl=file:///mnt/sol; licMolpro=null; }`. An idea what's wrong there?
```

markuskowa:
I suspect this that this is connected to the sandbox. You are not allowed to access arbitrary paths from your file system when building a derivation. Try adding '/mnt/sol' to `extra-sandbox-paths`in your `nix.conf`.

PR #46, opened 2021-05-19
sharc: fix molpro passing

sheepforce:
Fixes also the other sharc versions, that didn't may get passed molpro attributes, that are `null`.


pkgs/apps/sharc/default.nix
```diff
@@ -1,6 +1,5 @@
 { stdenv, lib, fetchFromGitHub, makeWrapper, gfortran
-, openblasCompat, fftw, python2, molcas, molpro
-, useMolpro ? false
+, openblasCompat, fftw, python2, molcas, molpro ? null
```
markuskowa, line 2:
```suggestion
, blas, fftw, python2, molcas, molpro ? null
```

markuskowa:
Thanks!

PR #48, opened 2021-05-19
xcfun: simplify cmake part

markuskowa:
I simplified the cmake part and still builds. @sheepforce can you check if it breaks something?


markuskowa:
Btw.: this is a clean and simple derivation now. You could consider submitting it as an upstream derivation.

PR #49, opened 2021-05-20
Sharc Fixes

### LINKED ISSUE #47: Sharc v21
Hiho, 

for sharcV21 two issues arose: 

1) 
The variable $SHARC pointing towards the relevant python skripts has to be set. 

2) 
Sharc does not work with Openmolcas 21. 
In general, it seems to only be designed to work for versions 18.x 
There are two options:
i) pair it with Openmolcas 18.x
ii) allow read-write manipulations of the SHARC_MOLCAS.py to circumvent this restriction manually. However, it has to be seen if this would cause actual problems with Sharc. 

Thanks a lot

sheepforce:
This PR readds the surprisingly young version 2.1.1 (the tag exists now again ...) of SHARC, adds formal compatibility with current OpenMolcas and exports the `$SHARC` environment by means of a setupHook. Closes #47 

There is also OpenMolcas 18.09 now, which is the officially supported version in SHARC.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -161,6 +161,8 @@ let
 
       molcas = self.molcas2102;
 
+      molcas1809 = callPackage ./pkgs/apps/openmolcas/v18.09.nix { };
+
       molcas2010 = callPackage ./pkgs/apps/openmolcas/v20.10.nix { };
 
       molcas2102 = callPackage ./pkgs/apps/openmolcas/v21.02.nix { };
--- a/pkgs/apps/openmolcas/v18.09.nix
+++ b/pkgs/apps/openmolcas/v18.09.nix
@@ -1,0 +1,152 @@
+{ lib, stdenv, pkgs, fetchFromGitLab, fetchpatch, cmake, gfortran, perl
+, blas-i8, hdf5-full, python3, texlive, symlinkJoin
+, armadillo, makeWrapper, fetchFromGitHub, chemps2
+} :
+
+assert
+  lib.asserts.assertMsg
+  (blas-i8.isILP64 || blas-i8.passthru.implementation == "mkl")
+  "A 64 int integer BLAS implementation is required.";
+
+assert
+  lib.asserts.assertMsg
+  (builtins.elem blas-i8.passthru.implementation [ "openblas" "mkl" ])
+  "OpenMolcas requires OpenBLAS or MKL.";
+
+let
+  version = "18.09-02.09.2018";
+
+  python = python3.withPackages (ps : with ps; [ six pyparsing ]);
+
+  srcLibwfa = fetchFromGitHub {
+    owner = "libwfa";
+    repo = "libwfa";
+    rev = "efd3d5bafd403f945e3ea5bee17d43e150ef78b2";
+    sha256 = "0qzs8s0pjrda7icws3f1a55rklfw7b94468ym5zsgp86ikjf2rlz";
+  };
+
+in stdenv.mkDerivation {
+  pname = "openmolcas";
+  inherit version;
+
+  src = fetchFromGitLab {
+    owner = "Molcas";
+    repo = "OpenMolcas";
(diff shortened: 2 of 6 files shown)


default.nix
```diff
@@ -161,6 +161,8 @@ let
 
       molcas = self.molcas2102;
 
+      molcas1809 = callPackage ./pkgs/apps/openmolcas/v18.09.nix { };
```
markuskowa, line 164: Do we really need molcas-18.09 if it also runs with a newer version?

pkgs.nix
```diff
@@ -0,0 +1,42 @@
+let
```
markuskowa, line 1: I guess this was an accident?

pkgs/apps/sharc/21.nix
```diff
@@ -91,6 +93,10 @@ in stdenv.mkDerivation {
     done
   '';
 
+  setupHooks = [
```
markuskowa, line 96: Does that work with `nix-shell -p sharc`?

markuskowa (changes requested):

pkgs.nix
sheepforce: Ups, yes indeed. Is removed :grimacing:

default.nix
sheepforce: Hm, I guess we don't strictly need it. I would still keep it. I guess it is a good option to have the officially supported version available.

pkgs/apps/sharc/21.nix
sheepforce: Actually yes! (Very much to my surprise) :smile: (So with `nix-shell -p qchem.sharcV21`)

default.nix
markuskowa: OK, then let's keep it for now.

pkgs/apps/openmolcas/v18.09.nix
```diff
@@ -0,0 +1,152 @@
+{ lib, stdenv, pkgs, fetchFromGitLab, fetchpatch, cmake, gfortran, perl
+, blas-i8, hdf5-full, python3, texlive, symlinkJoin
+, armadillo, makeWrapper, fetchFromGitHub, chemps2
+} :
+
+assert
+  lib.asserts.assertMsg
+  (blas-i8.isILP64 || blas-i8.passthru.implementation == "mkl")
+  "A 64 bint integer BLAS implementation is required.";
```
markuskowa, line 9:
```suggestion
  "A 64 bit integer BLAS implementation is required.";
```

pkgs/apps/sharc/21.nix
```diff
@@ -75,7 +77,7 @@ in stdenv.mkDerivation {
                      --set-default MOLCAS ${molcas} \
                      --set-default BAGEL ${bagel} \
                      ${lib.optionalString (molpro != null) "--set-default MOLPRO ${molpro}/bin"} \
-                     ${lib.optionalString (orca != null) "--set-default ORCA ${orca}/bin"} \
+                     ${lib.optionalString (orca != null) "--set-default ORCADIR ${orca}/bin"} \
                      ${lib.optionalString (turbomole != null) "--set-default TURBOMOLE ${turbomole}/bin"} \
                      ${lib.optionalString (gaussian != null) "--set-default GAUSSIAN ${turbomole}/bin"}
```
markuskowa, line 82:
```suggestion
                     ${lib.optionalString (gaussian != null) "--set-default GAUSSIAN ${gaussian}/bin"}
```

pkgs/apps/sharc/21.nix
markuskowa: Nice

markuskowa:
Can you please squash the commits a little (down to 2 or 3)?  Then I'll merge.

sheepforce:
Done! :slightly_smiling_face:

markuskowa:
Thanks for the fixes!

Issue #52, opened 2021-05-22
AVX512 in openblas is broken


markuskowa:
## Problem
After re-building all packages on Xeon Scalable Bronze, which has AVX512, I noticed that multiple packages, which are linked against openblas, do not pass their tests anymore or their test never finish.
The affected packages that I have observed are:
* molcas (`tests.molcas`)
* chemps2 (see also https://github.com/markuskowa/NixOS-QChem/issues/40)
* pyscf (see also https://github.com/markuskowa/NixOS-QChem/issues/29)
* octave (the full feature override in the overlay)
* nwchem (`tests.nwchem`)

## Solution
Possible workarounds are:
* set `OPENBLAS_CORETYPE` at runtime (`HASWELL`, `ATHLON` work, `SKYLAKEX` fails). Needs to be set in every derivation!
* override openblas and add `NO_AVX512=1` to `makeFlags`.

openblas 0.3.15 is already in staging (need to check if the problem still exists here).

## Branches affected
* NixOS-QChem master
* nixpkgs unstable, release-21.05 (as of writing this issue)

cc @sheepforce 

EDIT: `openblas-0.3.15`, which is staging does not seem to solve any of the problems.

sheepforce:
Ah to bad that this affects even more packages. At the moment I am setting `OPENBLAS_CORETYPE=HASWELL` in the `preOverlay` and this works, but I guess `NO_AVX512=1` is more specific. Do we know if this problem also exists upstream at OpenBLAS itself, outside of NixPkgs?

markuskowa:
I have not found any other reports of bugs so far. Neither Debian nor Fedora turn it off in their builds. I am a bit puzzled that no one else has discovered this problem.
It could be that we have a fringe case here: AVX512 is not very widely used (only in more expensive CPUs) combined with the fact that HPC applications tend to use MKL more frequently than openblas.

sheepforce:
Hm, possibly we should let the OpenBLAS developers know?

Now that you've merged #54 we can maybe check if the problems in CP2K, XTB and Bagel with BLAS can be solved. I've built them all on my AVX512 machine an it is very well possible, that I've ran into such problems there.

markuskowa:
> Hm, possibly we should let the OpenBLAS developers know?

Yes that would be good, but this bug report will be very unspecific from the point of the developers: we have no idea which routines are affected (I am not sure yet how to write that but report).

[...]

PR #54, opened 2021-05-22
override openblas to fix AVX512 bug

markuskowa:
Workaround until the problem is fixed in upstream nixpkgs/openblas.
see https://github.com/markuskowa/NixOS-QChem/issues/52

CC @sheepforce

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -52,6 +52,13 @@ let
       # Upstream overrides
       #
 
+      # FIXME: workaround, remove when upstream openblas is fixed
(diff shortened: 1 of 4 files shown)


sheepforce:
Looks good to me until we have a valid fix upstream. I've been looking in the [OpenBLAS issues for something similar](https://github.com/xianyi/OpenBLAS/issues?q=is%3Aissue+AVX512+) and there seem to be many issues with Skylake CPUs and vector extensions, although I can't find something exactly as our problem.

PR #55, opened 2021-05-22
sharc: remove V1

markuskowa:
The sources of the package are not publicly available anymore. I would suggest to remove the package and focus the maintaince effort on version 2. Does anyone still use V1?

cc @sheepforce


sheepforce:
I am perfectly fine with removing it! We use it on a regular basis and even have a SHARC developer here, but no one is using v1.

Issue #56, opened 2021-05-23
Packages to be moved to upstream nixpkgs


markuskowa:
This is a list to keep track of packages that are either good enough to be submitted to upstream nixpkgs or can be easily brought into shape for upstream nixpkgs.

* [x] `f2c`
* [x] `cp2k`, may require a little bit of cleanup
* [x] `dkh`
* [x] `ergoscf`
* [x] `i-Pi`
* [x] `libint`
* [x] `libvori`
* [x] `pcmsolver`
* [x] `pyscf`
* [x] `qcelemental`
* [x] `qcengine`
* [x] `mvapich`
* [x] `xcfun`

sheepforce:
As discussed some time ago I've opened a PR against NixPkgs to be added as a maintainer and would take of a few of those. See https://github.com/NixOS/nixpkgs/pull/124370

markuskowa:
* NWchem is stable but the build system is too messy. I rather keep it local.
* XTB takes proprietary packages as inputs, making it less suitable for nixpkgs.
* GPAW could be doable but also requires `libvdwxc`.

sheepforce:
CP2K is running just fine now, I suppose. I would be fine with maintaining it upstream, if that's OK for you (I also use it regularly)? Upstream there is no bats test. Is it OK for upstream to simply inline them as a installCheckPhase?

markuskowa:
> CP2K is running just fine now, I suppose. I would be fine with maintaining it in NixOS/nixpkgs, if that's OK for you (I also use it regularly)? 

Yes, we can do that but `libvori` is missing upstream. We should PR that one first.

> Upstream there is no bats test. Is it OK for upstream to simply inline them as a installCheckPhase?

That is OK. We just keep the bats tests here. Expensive install checks are not so welcome upstream anyway since Hydra is already very busy. The advantage of having it here is that they are run local hardware (at least from my perspective).

sheepforce:
I was starting to port CP2K to nixpkgs and just noticed that also libint is missing. Would you like to bring it to nixpkgs or should I take care of this?

sheepforce:
> I was starting to port CP2K to nixpkgs and just noticed that also libint is missing. Would you like to bring it to nixpkgs or should I take care of this?

I have a branch [here](https://github.com/sheepforce/nixpkgs/tree/libint).

markuskowa:
I will take a look `libint`. That one requires some cleanup.

sheepforce:
[...]

PR #61, opened 2021-05-26
MKL -> OpenBlas wherever possible

sheepforce:
After #54 has been merged OpenBlas now actually works in more cases. I was able to replace MKL in Bagel, XTB and CP2K without affecting our test cases. For Bagel I've disabled LibXC support. It doesn't find it anymore (no idea why) but I guess as DFT in Bagel is not documented and not part of the public release anyway (and Bagel is also not the program I would choose to do DFT :smile: ), this should be fine.

Maybe good enough for those 3 for #56 ?

One package that I could not migrate is CREST, [as it explicitly calls mkl routines](https://github.com/grimme-lab/crest/blob/594ef26f266ec0366e4340a349351a7c7c2de0c0/src/solve.f#L266). :disappointed:

### DIFF
--- a/pkgs/apps/bagel/default.nix
+++ b/pkgs/apps/bagel/default.nix
@@ -1,23 +1,22 @@
 { stdenv, lib, fetchFromGitHub, autoconf, automake, libtool
 , makeWrapper, openssh
-, python, boost, mpi, libxc, blas
-, withScalapack ? false
-, scalapack
+, python3, boost, mpi, blas, lapack
+, scalapack ? null
+, optAVX ? true
 } :
 
 let
-  version = "1.2.2";
-
-  blasName = (builtins.parseDrvName blas.name).name;
-
-  mpiName = (builtins.parseDrvName mpi.name).name;
+  mpiName = mpi.pname;
   mpiType = if mpiName == "openmpi" then mpiName
        else if mpiName == "mpich"  then "mvapich"
        else if mpiName == "mvapich"  then mpiName
        else throw "mpi type ${mpiName} not supported";
 
-in stdenv.mkDerivation {
-  name = "bagel-${version}" + lib.optionalString (blasName != "openblas") "-${blasName}";
+  useMKL = blas.passthru.implementation == "mkl" && lapack.passthru.implementation == "mkl";
+
+in stdenv.mkDerivation rec {
+  pname = "bagel";
+  version = "1.2.2";
 
   src = fetchFromGitHub {
     owner = "nubakery";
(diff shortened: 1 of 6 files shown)


pkgs/apps/bagel/default.nix
```diff
@@ -27,25 +24,28 @@ in stdenv.mkDerivation {
...
-           + lib.optionalString (blasName == "mkl") "-L${blas}/lib/intel64";
-
-  LDFLAGS = lib.optionalString (blasName == "mkl") "-L${blas}/lib/intel64";
+  CXXFLAGS = builtins.toString ([
+    "-DNDEBUG"
+    "-O3"
+    "-DCOMPILE_J_ORB"
+    "-lblas"
+    "-llapack"
+  ] ++ lib.lists.optional optAVX "-mavx"
```
markuskowa, line 39:
```suggestion
  ] ++ lib.lists.optional optAVX "-mavx -mavx2"
```

markuskowa:
> One package that I could not migrate is CREST, as it explicitly calls mkl routines

That's OK. MKL has more than just BLAS and LAPACK. Some packages will just be explicit MKL tied packages.

The bagel derivation is much cleaner now. However, the tests do finish when I try to build them on my AMD nodes now (same for `qchem.tests.bagel` and `qchem-mkl.tests.bagel`) but it finished in a reasonable time on the Intel node!?  I am not sure what could cause this. Bagel is developed with MKL in mind - we may need to keep it that way for now.

sheepforce:
Sorry, I didn't get the BAGEL testing issues completely. The tests take an unreasonable time on AMD hardware but much shorter on Intel? Even if you build with MKL as `blasProvider` and `lapackProvider`? This sounds terrible :( `qchem.tests.bagel` took about 10-15 minutes on my 4 year old Intel i5 quad core at home.
There are quite a few places, where MKL routines are explicitly called in the BAGEL code, yes. On our Intel cluster I've been using BAGEL with OpenBlas with some success, though. 

Before we are dropping OpenBlas here again, I would prefer to unwrap the MKL, if it is set as `blasProvider`, to allow us to keep a open version of BAGEL, which we can cache. BAGEL takes just so much time to compile.

markuskowa:
> The tests take an unreasonable time on AMD hardware but much shorter on Intel? Even if you build with MKL as blasProvider and lapackProvider? This sounds terrible

Yes, some tests are hanging on the AMD nodes (different ones for MKL blas/lapackProvider and openblas). 

> Before we are dropping OpenBlas here again, I would prefer to unwrap the MKL, if it is set as blasProvider, to allow us to keep a open version of BAGEL, which we can cache. BAGEL takes just so much time to compile.

OK, go ahead.

sheepforce:
Done! :slightly_smiling_face:

markuskowa:
OK, but with MKL we have used the MKL packaged scalapack library. Now it always uses the standard scalapack package.
Also: let's make Bagel with MKL the default build. This is how it is intended (and tested) by the developers.

Btw. one of the test that takes quite long is `hf_svp_b3lyp.json`, that may be linked to removal of `libxc`. Does it still work if you use `libxc4`.

markuskowa:
OK, I walk back on my last post:
* Let's just remove the faulty test by skipping it (it is `hf_svp_b3lyp_opt.json`). I have no problems when the test is removed.
* otherwise keep as is right now: no libxc, openblas default, MKL unwrapped, standard Scalapack.

sheepforce:
Alright, has been done like this. :slightly_smiling_face: While mixing OpenBlas-Scalapack with MKL blas is for sure strange, I've never had any problems with combinations like this, actually.

PR #62, opened 2021-05-28
cp2k: 8.1.0 -> 8.2.0

sheepforce:
CP2K 8.2 released basically just now. This is the update and supports libxc 5. (Does anything else use libxc4 or can we drop it now?)

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -128,7 +128,6 @@ let
       chemps2 = callPackage ./pkgs/apps/chemps2 { };
 
       cp2k = callPackage ./pkgs/apps/cp2k {
-        libxc = self.libxc4;  # patches are are required for libxc5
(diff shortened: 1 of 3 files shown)


markuskowa:
> (Does anything else use libxc4 or can we drop it now?)

libxc4 can be removed now. It was only there for cp2k-8.1.0.

sheepforce:
Removed :slightly_smiling_face:

markuskowa:
Oops. I overlooked that `psi4` still uses `libxc4`

sheepforce:
Haha, I wasn't even aware anymore myself. :smile:  OK, then we keep libxc for now. Psi4 1.4 or 1.5 will probably port to libint2 and libxc 5.

PR #64, opened 2021-05-31
VMD Binaries with CUDA support

sheepforce:
Adds the distributed prebuilt VMD binaries. They have SSE, AVX and CUDA enabled and offer additional plugins and Tachyon functionality. A flag has been introduced to enable CUDA for the overlay, similar to `optAVX`. In principle we could also use this to enable CUDA in e.g. CP2K, GPAW, Octopus and Gaussian (although from my experience with CUDA besides VMD this is usually not worth the effort).
While VMD and the missing plugins now run (also in memory ray-tracing), the Nvidia OSPrayRendering does not. It simply crashes with an `Illegal Instruction` error. I've run this version of VMD on my machine already under OpenSuse, where it worked, so this is definitely not a hardware issue. I suspect the issue to be somewhere in the Nvidia libraries, but I am not too familiar with the internal of NixOS graphics stack.

### DIFF
--- a/pkgs/apps/vmd/binary.nix
+++ b/pkgs/apps/vmd/binary.nix
@@ -1,0 +1,93 @@
+{ stdenv, lib, requireFile, makeWrapper, writeScriptBin, bash, perl, tcl-8_5, tk-8_5
+, netcdf, libGLU, xorg, fltk, vrpn, flex, bison, libGL_driver, cudatoolkit, autoPatchelfHook
+}:
+assert
+  lib.asserts.assertMsg
+  (stdenv.isLinux && stdenv.isx86_64)
+  "The VMD binaries require an x86_64 linux OS with CUDA support.";
+
+let homepage = "https://www.ks.uiuc.edu/Research/vmd/";
+
+in stdenv.mkDerivation rec {
+  pname = "vmd";
+  version = "1.9.3";
+
+  src = requireFile {
+    url = homepage;
+    name = "vmd-1.9.3.bin.LINUXAMD64-CUDA8-OptiX4-OSPRay111p1.opengl.tar.gz";
+    sha256 = "9427a7acb1c7809525f70f635bceeb7eff8e7574e7e3565d6f71f3d6ce405a71";
+  };
+
+  nativeBuildInputs = [
+    perl
+    makeWrapper
+    autoPatchelfHook
+  ];
+
+  buildInputs = [
+    libGLU
+    xorg.libX11
+    xorg.libXinerama
+    xorg.libXi
+    tcl-8_5
+    tk-8_5
+    netcdf
+    fltk
+    vrpn
+    flex
+    bison
+    libGL_driver
+    cudatoolkit.out
+    cudatoolkit.lib
+    # nvidia_x11
+  ];
+
+  postPatch = ''
+    substituteInPlace ./configure \
+      --replace '/usr/local' "$out" \
+      --replace '-ll' '-lfl'
+
+    patchShebangs ./configure
+  '';
+
+  sourceRoot = "vmd-${version}";
+
+  # non-standard configure script
+  configurePhase = ''
+    ./configure
+  '';
+
+  dontBuild = true;
+
+  preInstall = ''
+    cd src
+  '';
+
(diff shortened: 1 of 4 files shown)


pkgs/apps/vmd/binary.nix
```diff
@@ -0,0 +1,97 @@
+{ stdenv, lib, requireFile, makeWrapper, writeScriptBin, bash, perl, tcl-8_5, tk-8_5, symlinkJoin
+, netcdf, libGLU, xorg, fltk, vrpn, flex, bison, libGL_driver, cudatoolkit, autoPatchelfHook
+}:
+assert
+  lib.asserts.assertMsg
+  (stdenv.isLinux && stdenv.isx86_64)
+  "The VMD binaries with require an x86_64 linux OS with CUDA support.";
```
markuskowa, line 7:
```suggestion
  "The VMD binaries require an x86_64 linux OS with CUDA support.";
```

pkgs/apps/vmd/binary.nix
```diff
@@ -0,0 +1,97 @@
...
+
+let
+  homepage = "https://www.ks.uiuc.edu/Research/vmd/";
+  cudatoolkit_joined = symlinkJoin {
```
markuskowa, line 11: This doesn't work with the separate inputs directly in `buildInputs`?

pkgs/apps/vmd/binary.nix
```diff
@@ -0,0 +1,97 @@
...
+    name = "cudatoolkit_joined";
+    paths = [ cudatoolkit.out cudatoolkit.lib ];
+  };
+
+in stdenv.mkDerivation rec {
+  pname = "vmd";
+  version = "1.9.3";
+
+  src = requireFile rec {
```
markuskowa, line 20:
```suggestion
  src = requireFile {
```

pkgs/apps/vmd/binary.nix
```diff
@@ -0,0 +1,97 @@
...
+  '';
+
+  # libnvcuvid.so depends on the linuxPackages.nvidia_x11.
+  # This might be overriden in configuration.nix and should be detected at runtime at
+  # /var/run/opengl-driver/lib
+  autoPatchelfIgnoreMissingDeps = true;
+
+  postFixup = ''
+    wrapProgram $out/bin/vmd \
+      --set "LC_ALL" "C"
```
markuskowa, line 85: What is that required for? Should we really override local LC settings?

markuskowa (changes requested):

pkgs/apps/vmd/binary.nix
sheepforce: VMD tends to use system locales for its parser. In consequence, if you are on a system with a locale, that uses comma as decimal separator (like german aka it's a problem specifically for me :smile: ), the VMD parsers will fail to parse anything, as they suddenly expect xyz, molden, cube, pdb, ... files to use comma as a separator.

pkgs/apps/vmd/binary.nix
sheepforce: It does :)

sheepforce:
The interactive raytracing works now. I am not sure if somehow the changes in the cudatoolkit made a difference or if it is due to another computer now. At least I don't have the `Illegal Instruction` errors here.

pkgs/apps/vmd/binary.nix
markuskowa: OK, I see.

markuskowa:
The main issue with any OpenGL (or GPU related) app is that it needs to access the local OpenGL drivers, making the app inherently impure. On a NixOS the respective libs are in `/run/opengl-driver`. Not sure how the OpenGL issue is solved on a non-NixOS running nix but in the end the app from the nix store will also access your local OpenGL drivers from the OS.
The result is that bugs like `Illegal Instruction` may then be connected to the run-time environment and the OS.

markuskowa (approved):

sheepforce:
On non-NixOS there is [NixGL](https://github.com/guibou/nixGL), which I've used with some success on OpenSuse.

PR #72, opened 2021-06-09
meep: init at 1.18.0

sheepforce:
This adds the meep package for electromagnetics simulations as requested by my colleague @FabianGD. The Scheme interface is disabled for now, as meep somehow does not find `libctl` correctly. The python interface should work nonetheless.

@FabianGD please have a look if it behaves as intended :slightly_smiling_face:

### DIFF
--- a/pkgs/apps/meep/default.nix
+++ b/pkgs/apps/meep/default.nix
@@ -1,0 +1,103 @@
+{ stdenv, lib, buildPythonPackage, fetchFromGitHub, autoreconfHook, pkg-config
+, gfortran, mpi, blas, lapack, fftw, hdf5-full, swig, gsl, harminv, libctl
+, libGDSII, openssh, guile
+# Python
+, python, numpy, scipy, matplotlib, h5py, cython, autograd, mpi4py
+} :
+
+buildPythonPackage rec {
+  pname = "meep";
+  version = "1.18.0";
+
+  src = fetchFromGitHub {
+    owner = "NanoComp";
+    repo = pname;
+    rev = "v${version}";
+    sha256= "08l4mczkh08dp90c4dlnyx6lsg093li0xqnnbh7q8k363cr8lx81";
+  };
+
+  nativeBuildInputs = [
+    autoreconfHook
+    gfortran
+    pkg-config
+    swig
+    mpi
+  ];
+
+  buildInputs = [
+    gsl
+    blas
+    lapack
+    fftw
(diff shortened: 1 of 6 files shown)


sheepforce:
@markuskowa should meep be moved into the `python3` set?

markuskowa:
> should meep be moved into the python3 set?

Because it also installs a python interface? The derivation is not a python package in the first place, so I would say it is good as it is. You could create a separate, simplified derivation (if possible) that only install the python interface as a proper python package. I am not a 100% sure if that is the best solution here, though. Take a look at `nixpkgs` and see how such cases are solved there.

sheepforce:
OK, works and ready for feedback @markuskowa :slightly_smiling_face: Thanks to @FabianGD the python interface now works properly. 

As it is mainly used through the Python interface, I've put it into the the `python3.pkgs`. Somehow similar as it is done with Psi4.

pkgs/apps/meep/default.nix
```diff
@@ -0,0 +1,103 @@
...
+  dontUsePipInstall = true;
+  dontUseSetuptoolsCheck = true;
+
+  HDF5_MPI = "ON";
+  PYTHON = "${python}/bin/${python.executable}";
+
+  enableParallelBuilding = true;
+
+  configureFlags = [
+    "--without-libctl"
```
markuskowa, line 63: Is that intended? `libctl` is in the `buildInputs`.

pkgs/apps/meep/default.nix
```diff
@@ -0,0 +1,103 @@
...
+    rev = "v${version}";
+    sha256= "08l4mczkh08dp90c4dlnyx6lsg093li0xqnnbh7q8k363cr8lx81";
+  };
+
+  nativeBuildInputs = [
+    autoreconfHook
+    gfortran
+    pkg-config
+    swig
+    mpi
```
markuskowa, line 24: It should be enough to have `mpi` in `propagatedBuildInputs`.

sheepforce: Interestingly it is not. I'm not sure why, but `configure` does not find it, if only in `propagatedBuildInputs`

pkgs/apps/meep/default.nix
sheepforce: Yep, it is still needed and wont work without. So one part of `configure` finds it, while another one is not finding it ...

PR #77, opened 2021-06-21
gmultiwfn: init at 3.4.1-0

FurstPlattner:
Adding gMultiwfn attribute, as standard Multiwfn encounters issues when dealing with large wave functions. cc @sheepforce

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -176,6 +176,8 @@ let
 
       multiwfn = callPackage ./pkgs/apps/multiwfn { };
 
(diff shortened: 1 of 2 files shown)


sheepforce:
Can confirm segfaults for "many" (lets say somewhere around 1500) basis functions. Unfortunately this problem is known since some time and similar to the ifort restrictions not fixed. gMultiwfn is the only workaround I am aware of, even if it is an old version.

PR #78, opened 2021-06-23
pysisyphus: 0.7 -> 0.7.post1

sheepforce:
@markuskowa could you please cherrypick this also for the 21.05 branch?

### DIFF
--- a/pkgs/apps/pysisyphus/default.nix
+++ b/pkgs/apps/pysisyphus/default.nix
@@ -57,7 +57,7 @@ let
 in
   buildPythonPackage rec {
     pname = "pysisyphus";
-    version = "0.7";
+    version = "0.7.post1";
 
     nativeBuildInputs = [ makeWrapper ];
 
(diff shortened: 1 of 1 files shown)


sheepforce:
Just updated nixpkgs to b27eaa18b47b77fbbf9f3bd7a17046b2d64b637b and it does not build due to errors in `python3.pkgs.dask`, which does not pass its test suite. Happens on my local machine and on our Hydra. It's the first time that dask is built from source and not coming from `cache.nixos.org`. Do you have the same problem?

sheepforce:
Looking into this a little bit reveals that those are some file system access tests, that are again done with fsspec. There were multiple issues already with fsspec and its file system dependency (https://github.com/NixOS/nixpkgs/pull/99178, https://github.com/NixOS/nixpkgs/pull/99094, https://github.com/NixOS/nixpkgs/pull/100153 and especially https://github.com/NixOS/nixpkgs/pull/99499). Possibly I am hitting a problem here again with my BTRFS file systems :neutral_face: 
The problem possibly solves itself when dask is cached from hydra.

markuskowa:
Yes, I have the same problem. It means that dask is broken in nixpkgs. It is probably not related to your FS.

sheepforce:
Alright. So should we wait with the merge for this upstream fix? In the end it is not really related to Pysisyphus and 0.7 will also be broken by Dask.

markuskowa:
Should already be fixed on master. I'll wait until it hits nixpkgs-unstable.

markuskowa:
Back-ported to 21.05 in 131b91cc3cd687e3169c29e54e5a31b8e7928117

PR #79, opened 2021-06-23
meep: fixing mpi runtime

sheepforce:
@FabianGD just discovered runtime problems with MPI parallel Meep execution. This affects writing HDF5-files mainly. This PR fixes this.

### DIFF
--- a/pkgs/apps/meep/default.nix
+++ b/pkgs/apps/meep/default.nix
@@ -1,8 +1,8 @@
 { stdenv, lib, buildPythonPackage, fetchFromGitHub, autoreconfHook, pkg-config
-, gfortran, mpi, blas, lapack, fftw, hdf5-full, swig, gsl, harminv, libctl
+, gfortran, mpi, blas, lapack, fftw, hdf5-mpi, swig, gsl, harminv, libctl
 , libGDSII, openssh, guile
 # Python
-, python, numpy, scipy, matplotlib, h5py, cython, autograd, mpi4py
+, python, numpy, scipy, matplotlib, h5py-mpi, cython, autograd, mpi4py
 } :
 
 buildPythonPackage rec {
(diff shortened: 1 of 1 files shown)


pkgs/apps/meep/default.nix
```diff
@@ -71,27 +71,41 @@ buildPythonPackage rec {
...
-    sim = mp.Simulation(cell_size=cell,
-                    boundary_layers=pml_layers,
-                    geometry=geometry,
-                    sources=sources,
-                    resolution=resolution)
-    sim.run(until=200)
-    EOF
-  '';
+  checkPhase =
+    let testInput = writeTextFile {
```
markuskowa, line 75: I would rather keep the
```
cat  > test.py << EOF
...
EOF
```
construction here. That should be faster than building the text file.

pkgs/apps/meep/default.nix
```diff
@@ -71,27 +71,41 @@ buildPythonPackage rec {
...
+    in ''
+      export PYTHONPATH="$out/lib/${python.libPrefix}/site-packages:$PYTHONPATH"
+
+      export OMP_NUM_THREADS=1
+
+      # Fix to make mpich run in a sandbox
+      export HYDRA_IFACE=lo
+      export OMPI_MCA_rmaps_base_oversubscribe=1
+
+      cat ${testInput}
```
markuskowa, line 105: That was for debugging purposes?

pkgs/apps/meep/default.nix
sheepforce: OK, can change it again. I've removed the original version as only one of the launched interpreters got the input. But of course we can use cat as you suggest.

pkgs/apps/meep/default.nix
sheepforce: It was. missed it again, sorry.

markuskowa (approved):

PR #80, opened 2021-06-24
gpaw: GPAW_SETUP_PATH and 21.1.0 -> 21.6.0

sheepforce:
GPAW always requires the `GPAW_SETUP_PATH` variable, similar to how Sharc requires `$SHARC`. This uses the same mechanism using `setupHooks` to make sure, that not only the `gpaw` executable (as in the previous version) sees this variable, but also the Python interface works without manual intervention.

### DIFF
--- a/pkgs/apps/gpaw/SetupPath.patch
+++ b/pkgs/apps/gpaw/SetupPath.patch
@@ -1,0 +1,18 @@
+diff --git a/gpaw/__init__.py b/gpaw/__init__.py
+index b5c029e13..518c16b13 100644
+--- a/gpaw/__init__.py
++++ b/gpaw/__init__.py
+@@ -201,12 +201,7 @@ def initialize_data_paths():
+     try:
+         setup_paths[:0] = os.environ['GPAW_SETUP_PATH'].split(os.pathsep)
+     except KeyError:
+-        if len(setup_paths) == 0:
+-            if os.pathsep == ';':
+-                setup_paths[:] = [r'C:\gpaw-setups']
+-            else:
+-                setup_paths[:] = ['/usr/local/share/gpaw-setups',
+-                                  '/usr/share/gpaw-setups']
++        setup_paths[:0] = ["@gpawSetupPath@"]
+ 
+ 
+ read_rc_file()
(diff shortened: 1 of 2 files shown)


FabianGD:
I was playing around with the changes here and while using `qchem.gpaw` (as the Python Application) correctly sets the `GPAW_SETUP_PATH` environment variable, this for some reason does not work for using it as a Python package (via building a Python env as `python3.withPackages(ps: with ps; [qchem.python3.pkgs.gpaw])`). 
In this case, the variable does not seem to be set in the resulting shell. 

This might be an user error (I'm quite new to nix), so is there some more "nix magic" hidden here? 😄

sheepforce:
OK, so turns out that the `setupHooks` are ignored when using them in the `python.withPackages` mechanism. The brute-force workaround here is to hardcode the `GPAW_SETUP_PATH` in the python module.

pkgs/apps/gpaw/SetupPath.patch
```diff
@@ -0,0 +1,21 @@
...
+-    try:
+-        setup_paths[:0] = os.environ['GPAW_SETUP_PATH'].split(os.pathsep)
+-    except KeyError:
+-        if len(setup_paths) == 0:
+-            if os.pathsep == ';':
+-                setup_paths[:] = [r'C:\gpaw-setups']
+-            else:
+-                setup_paths[:] = ['/usr/local/share/gpaw-setups',
+-                                  '/usr/share/gpaw-setups']
++    setup_paths[:0] = @gpawSetupPath@
```
markuskowa, line 18: Maybe a little less brute force? Somthing like this:
```suggestion
try:
    setup_paths[:0] = os.environ['GPAW_SETUP_PATH'].split(os.pathsep)
except KeyError:
        if len(setup_paths) == 0:
            setup_paths[:0] = @gpawSetupPath@
```

That way you can still use `GPAW_SETUP_PATH` as intended by gpaw, but you also have the built default.

markuskowa: We can then also remove the `preCheck` phase?

sheepforce: Yep, very good idea. Changed it.

PR #82, opened 2021-06-30
pysisyphus: make all qc packages optional, add openssh for dask

sheepforce:
In an attempt to build small self-contained archives with nix-bundle, I've made all quantum chemistry packages of pysis optional.
Furthermore Dask seems to require OpenSSH at runtime.


markuskowa:
Is the default version then still functional?

sheepforce:
Yes. In principle all qc engines are optional to Pysisyphus. They can still be used by pointing the `PYSISRC` variable to another configuration file. So Pysisyphus with all packages being null, can also use qc programs, that are not provided by Nix. Needed to use this on our cluster like this.

markuskowa:
OK, so you want a conscious opt-in rather than a default set.

sheepforce:
More or less; here, `callPackage` actually provides all the dependencies as non-null values and therefore builds a full pysisyphus derivation. But it is good to have a slim derivation for some purposes, where the closure is not 30 GB in total.

PR #91, opened 2021-08-03
dirac: init at 21.0

FabianGD:
This PR adds [DIRAC 21.0](https://diracprogram.org/) and it's dependency [ExaTensor](https://gitlab.com/DmitryLyakh/ExaTensor) to the collection. 

ExaTensor is currently in (more or less) early development so that no version tags exist. DIRAC requires a very specific version so I used the last 6 characters of the git hash as a reference. Is there a better way about that?

Furthermore, DIRAC has the possibility to include PCMSolver (at v1.1.1) to the build. I skip this for now, as fixing it requires even more patching of the CMake build system of DIRAC. What do you think about that?

Best,
Fabi

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -130,9 +130,13 @@ let
 
       dalton = callPackage ./pkgs/apps/dalton { };
 
-      dkh = callPackage ./pkgs/apps/dkh { };
-
       dftd3 = callPackage ./pkgs/apps/dft-d3 { };
+
+      dirac = callPackage ./pkgs/apps/dirac rec {
+        exatensor = self.exatensor;
+      };
+
+      dkh = callPackage ./pkgs/apps/dkh { };
 
       ergoscf = callPackage ./pkgs/apps/ergoscf { };
 
--- a/default.nix
+++ b/default.nix
@@ -136,6 +136,10 @@ let
 
       ergoscf = callPackage ./pkgs/apps/ergoscf { };
 
+      exatensor = callPackage ./pkgs/apps/exatensor rec {
+        mpi = super.mpi.override { gfortran = super.gfortran8; };
+      };
+
       gaussview = callPackage ./pkgs/apps/gaussview { };
 
       gdma = callPackage ./pkgs/apps/gdma { };
--- a/pkgs/apps/dirac/default.nix
+++ b/pkgs/apps/dirac/default.nix
@@ -1,0 +1,129 @@
+{ stdenv, lib, fetchurl, gfortran, cmake, makeWrapper
+, which, openssh, blas, lapack, mpi, boost, exatensor, python3
+} :
+
+assert
+  lib.asserts.assertMsg
+  (!blas.isILP64)
+  "A 32 bit integer implementation of BLAS is required.";
+
+stdenv.mkDerivation rec {
+  pname = "dirac";
+  version = "21.0";
+
+  nativeBuildInputs = [
+    which
+    gfortran
+    cmake
+    python3
+    makeWrapper
+    openssh
+  ];
+
+  # For now, DIRAC is built without the PCMSolver optional dependency.
+  buildInputs = [
+    blas
+    lapack
+    boost
+    exatensor
+  ];
+
+  propagatedBuildInputs = [
+    mpi
+  ];
+
+  src = fetchurl {
+    url = "https://zenodo.org/record/4836496/files/DIRAC-${version}-Source.tar.gz";
+    sha256 = "0vqmin24xqkcp3rvml0dimc4kjmb3m37c3l6bkjy4y6d3xyb41yg";
+  };
+
+  patches = [
+    ./0001-exatensor-cmake.patch
+  ];
+
+  postPatch = ''
+    substituteInPlace ./cmake/custom/exatensor.cmake \
+    --subst-var-by exatensor ${exatensor}
+
+    patchShebangs .
+  '';
+
+  FC = "mpif90";
+  CC = "mpicc";
+  CXX = "mpicxx";
+  MATH_ROOT = blas;
+  ENABLE_EXATENSOR = "ON";
+
+  /*
+  Cmake is required to build but adding it to the buildinputs then ignores the
+  setup script. Therefore i call the script here manually but cmake is invoked
+  by setup.
+  */
+
+  configurePhase = ''
+    ./setup --prefix=$out --mpi
+  '';
+
+  preBuild = ''
+    cd build
+  '';
+
+  # Parallel building of the DIRAC package introduces all kinds of unpredictable
+  # bugs when modules within a single folder depend on each other.
+  # enableParallelBuilding = true;
+
+  hardeningDisable = [ "format" ];
+  doInstallCheck = true;
+
+  /*
+  Make the MPI stuff available to the DIRAC script by hard-coding the MPI path.
+  Moving pam-dirac to pam to conform with the DIRAC manual. Usually there is a
+  weird hack in the install.cmake to install pam into the bin folder.
+
+  The pam script is just a Python script that sets a ton of ENV variables.
+  */
+  postFixup = ''
+    mv $out/bin/pam-dirac $out/bin/pam
+
+    substituteInPlace $out/bin/pam \
+      --replace "find_executable('mpirun')" "'${mpi}/bin/mpirun'"
+  '';
+
+  # This is a small initial calculation to see whether everything works just fine.
+  # http://diracprogram.org/doc/release-21/tutorials/getting_started.html
+  installCheckPhase = ''
+    cat > methanol.xyz << EOF
+      6
+      my first DIRAC calculation # anything can be in this line
+      C       0.000000000000       0.138569980000       0.355570700000
+      O       0.000000000000       0.187935770000      -1.074466460000
+      H       0.882876920000      -0.383123830000       0.697839450000
+      H      -0.882876940000      -0.383123830000       0.697839450000
+      H       0.000000000000       1.145042790000       0.750208830000
+      H       0.000000000000      -0.705300580000      -1.426986340000
+    EOF
+
+    cat > hf.inp <<-EOF
+    **DIRAC
+    .WAVE FUNCTION
+    **WAVE FUNCTION
+    .SCF
+    **MOLECULE
+    *BASIS
+    .DEFAULT
+    cc-pVDZ
+    *END OF INPUT
+    EOF
+
+    $out/bin/pam --mpi=2 --mol=$(pwd)/methanol.xyz --inp=$(pwd)/hf.inp --scratch=$(pwd)
+  '';
+
+  passthru = { inherit mpi; };
+
+  meta = with lib; {
+    description = "The DIRAC program computes molecular properties using relativistic quantum chemical methods.";
+    license = licenses.unfree;
+    homepage = "https://diracprogram.org/";
+    platforms = platforms.linux;
+  };
+}
--- a/pkgs/apps/exatensor/default.nix
+++ b/pkgs/apps/exatensor/default.nix
@@ -1,0 +1,51 @@
+{ stdenv, lib, fetchFromGitLab, blas, mpi } :
+
+let
+  version = "2020-07-15";
+
+in stdenv.mkDerivation rec {
+  pname = "exatensor";
+  inherit version;
+
+  src = fetchFromGitLab {
+    owner = "DmitryLyakh";
+    repo = pname;
+    rev = "d304c03b7289525f250245ece1b4a13a5973fef4";
+    sha256 = "0zlqwpdqfycwvys39lvvj94cprhaxmyaxxs1c66f10dazjbafp49";
+  };
+
+  buildInputs = [ blas.passthru.provider ];
+
+  MPILIB = if (mpi.pname == "openmpi") then "OPENMPI"
+    else if (mpi.pname == "mpich") then "MPICH"
+    else throw "Only openmpi and mpich supported by ${pname}.";
+
+  PATH_OPENMPI = mpi;
+  PATH_MPICH = mpi;
+
+  BLASLIB =
+    if (blas.passthru.implementation == "openblas") then "OPENBLAS"
+    else if (blas.passthru.implementation == "mkl") then "MKL"
+    else throw "Only MKL and OPENBLAS supported by ${pname}.";
+
+  PATH_BLAS_OPENBLAS = blas.passthru.provider;
+  PATH_BLAS_MKL = blas.passthru.provider;
+  PATH_BLAS_MKL_DEP = "${blas.passthru.provider}/lib";
+  PATH_BLAS_MKL_INC = "${blas.passthru.provider}/include";
+
+  installPhase = ''
+    mkdir -p $out/lib $out/bin $out/include
+    cp -r ./bin/* $out/bin
+    cp -r ./lib/* $out/lib
+    cp -r ./include/* $out/include
+  '';
+
+  meta = with lib; {
+    description = "ExaTENSOR is a basic numerical tensor algebra library for
+distributed HPC systems equipped with multicore CPU and NVIDIA or AMD GPU.";
+    homepage = https://gitlab.com/DmitryLyakh/ExaTensor;
+    license = licenses.lgpl3Only;
+    platforms = platforms.linux;
+  };
+}
+
--- a/pkgs/apps/dirac/0001-exatensor-cmake.patch
+++ b/pkgs/apps/dirac/0001-exatensor-cmake.patch
@@ -1,0 +1,192 @@
+diff --git a/cmake/custom/exatensor.cmake b/cmake/custom/exatensor.cmake
+index d203dff..40ca09e 100644
+--- a/cmake/custom/exatensor.cmake
++++ b/cmake/custom/exatensor.cmake
+@@ -1,140 +1,44 @@
++# Patched CMAKE file that does not download any exatensor.
++
+ #Check whether ExaTensor can be used
+-if(ENABLE_EXATENSOR)
+-   if(NOT ENABLE_MPI)
+-       message(STATUS "Using only the TALSH (serial) component of the ExaTensor library")
+-       set(EXATENSOR_TALSH EXA_TALSH_ONLY=YES)
+-       # The following weird construction is needed because these cmake variables are otherwise not active inside the environment and make that is called below
+-       set(TALSH_COMPILERS "CMAKE_Fortran_COMPILER=${CMAKE_Fortran_COMPILER} CMAKE_C_COMPILER=${CMAKE_C_COMPILER} CMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}")
+-   else()
+-       if (NOT (${MPI_Fortran_COMPILER} MATCHES "mpif90" OR ${MPI_Fortran_COMPILER} MATCHES "mpifort"))
+-          message(STATUS "This MPI compiler is not supported by ExaTENSOR, ExaCorr is switched off")
+-          set(ENABLE_EXATENSOR OFF)
+-       endif()
+-   endif()
+-   if(ENABLE_64BIT_INTEGERS)
+-       message(STATUS "ExaTensor library can not work with 64 bit integers, switched off")
+-       set(ENABLE_EXATENSOR OFF)
+-   endif()
++set(ENABLE_EXATENSOR ENABLED)
++
++# ExternalProject_Add seems required as later in the build process,
++# DIRAC includes exatensor via the CMAKE dependency system. It's not 100%
++# clear to me, so I'll just have that as a dummy here.
++ExternalProject_Add(exatensor
++    PREFIX exatensor-tmp
++    INSTALL_DIR @exatensor@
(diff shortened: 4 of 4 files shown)


default.nix
```diff
@@ -130,12 +130,23 @@ let
 
       dalton = callPackage ./pkgs/apps/dalton { };
 
+      dirac = callPackage ./pkgs/apps/dirac rec {
+        stdenv = super.gcc8Stdenv;
```
markuskowa, line 134: I don't think we need to override the c-compiler. Replacing gfortran is sufficient.

default.nix
```diff
@@ -130,12 +130,23 @@ let
...
+      };
+
       dkh = callPackage ./pkgs/apps/dkh { };
 
       dftd3 = callPackage ./pkgs/apps/dft-d3 { };
 
       ergoscf = callPackage ./pkgs/apps/ergoscf { };
 
+      exatensor = callPackage ./pkgs/apps/exatensor rec {
+        stdenv = super.gcc8Stdenv;
```
markuskowa, line 146: Same here: I don't think we need to override the c-compiler. Replacing gfortran is sufficient.

pkgs/apps/dirac/default.nix
```diff
@@ -0,0 +1,154 @@
...
+  buildInputs = [
+    blas
+    lapack
+    boost
+    exatensor
+  ];
+
+  propagatedBuildInputs = [
+    mpi
+    which
```
markuskowa, line 31: Which should be in `nativeBuildInputs`

pkgs/apps/dirac/default.nix
```diff
@@ -0,0 +1,154 @@
...
+  ];
+
+  src = fetchurl {
+    url = "https://zenodo.org/record/4836496/files/DIRAC-${version}-Source.tar.gz";
+    sha256 = "0vqmin24xqkcp3rvml0dimc4kjmb3m37c3l6bkjy4y6d3xyb41yg";
+  };
+
+  postPatch = ''
+    # cd $src
```
markuskowa, line 40:
```suggestion
```

pkgs/apps/dirac/default.nix
```diff
@@ -0,0 +1,154 @@
...
+    patchShebangs .
+
+    cat > cmake/custom/exatensor.cmake << EOF
```
markuskowa, line 43: Can you please provide this as an actual patch?

pkgs/apps/dirac/default.nix
```diff
@@ -0,0 +1,154 @@
...
+  ENABLE_EXATENSOR = "ON";
+
+  /*
+  Cmake is required to build but adding it to the buildinputs then ignores the
+  setup script. Therefore i call the script here manually but cmake is invoked
+  by setup.
+  */
+
+  configurePhase = ''
+    ./setup --prefix=$out --mpi && cd build
```
markuskowa, line 100: This is a nitpick, but `cd build` belongs into `preBuild`.

pkgs/apps/dirac/default.nix
```diff
@@ -0,0 +1,154 @@
...
+  doInstallCheck = true;
+
+  /*
+  Make the MPI stuff available to the DIRAC script. Direct exposure of MPI is not necessary.
+  */
+  postFixup = ''
+    wrapProgram $out/bin/pam-dirac \
+      --prefix PATH : ${mpi}/bin \
+      --prefix PATH : ${which}/bin \
+      --prefix PATH : ${openssh}/bin
```
markuskowa, line 114: I am not sure hard coding the openssh path is a good idea. The program should use the system provided ssh.
You can put `openssh` in `nativeBuildInputs` to ensure the `installCheckPhase` runs.

pkgs/apps/dirac/default.nix
```diff
@@ -0,0 +1,154 @@
...
+  enableParallelBuilding = true;
+  hardeningDisable = [ "format" ];
+  doInstallCheck = true;
+
+  /*
+  Make the MPI stuff available to the DIRAC script. Direct exposure of MPI is not necessary.
+  */
+  postFixup = ''
+    wrapProgram $out/bin/pam-dirac \
+      --prefix PATH : ${mpi}/bin \
```
markuskowa, line 112: Is `pam-dirac` just a script (not a binary)? If yes, you could use `substituteInPlace` to directly hard code the path instead of using a wrapper.

pkgs/apps/exatensor/default.nix
```diff
@@ -0,0 +1,54 @@
+{ stdenv, lib, fetchFromGitLab, blas, mpi } :
+
+let
+  version = "73fef4";
```
markuskowa, line 4: Please use the date of the commit as version number (formatted YYYY-MM-DD).

pkgs/apps/exatensor/default.nix
```diff
@@ -0,0 +1,54 @@
...
+
+in stdenv.mkDerivation rec {
+  pname = "exatensor";
+  name = "${pname}-${version}";
```
markuskowa, line 8:
```suggestion
  inherit version;
```

pkgs/apps/exatensor/default.nix
```diff
@@ -0,0 +1,54 @@
...
+  cp -r ./lib/* $out/lib
+  cp -r ./include/* $out/include
+  '';
+
+  meta = with lib; {
+    description = "ExaTENSOR is a basic numerical tensor algebra library for
+distributed HPC systems equipped with multicore CPU and NVIDIA or AMD GPU.";
+    homepage = https://gitlab.com/DmitryLyakh/ExaTensor;
+    license = licenses.lgpl3Only;
+    maintainers = [  ];
```
markuskowa, line 50:
```suggestion
```

default.nix
```diff
@@ -130,12 +130,23 @@ let
...
+        mpi = super.mpi.override { inherit stdenv; gfortran = super.gfortran8; };
```
markuskowa, line 147: I wonder if it is possible to just provide gfortran with a fixed version directly to exatensor (to avoid rebuilding openmpi)?
Does `gfortran11` work here?

markuskowa (changes requested):
Thanks for the contribution. I left a few comments.
Right now, I can't build the PR. It fails with:
```
/build/DIRAC-21.0-Source/external/stieltjes/stieltjesmod.f90:67:10:

       use stieltjes_ql_diag
          1
Fatal Error: Can't open module file 'stieltjes_ql_diag.mod' for reading at (1): No such file or directory
```

CC @sheepforce: do you have any comments or ideas how to improve the PR?

default.nix
sheepforce: ExaTensor explicitly states that it requires GCC 8 or 11 and it actually did not work with the normal `stdenv`

pkgs/apps/exatensor/default.nix
```diff
@@ -0,0 +1,54 @@
...
+  name = "${pname}-${version}";
+
+  src = fetchFromGitLab {
+    owner = "DmitryLyakh";
+    repo = pname;
+    rev = "d304c03b7289525f250245ece1b4a13a5973fef4";
+    sha256 = "0zlqwpdqfycwvys39lvvj94cprhaxmyaxxs1c66f10dazjbafp49";
+  };
+
+  # nativeBuildInputs = [ make ];
```
sheepforce, line 17: Basically never needed with `stdenv`
```suggestion
-  # nativeBuildInputs = [ make ];
```

default.nix
markuskowa: > It might work then again to give the normal stdenv to Dirac?

I would try that first. All other dependencies are also still  built with the default `stdenv`. 

Also: does `exatensor` work with `gcc11stdenv`? In that case we can drop this hack as soon as upstream upgrades to gcc11.

sheepforce: The ExaTensor stuff is unfortunately very hacky but required. There is a combination of the following problems, which make this the only viable option, i would guess:

- OpenMPI does not build with `gcc11stdenv` due to implicit typecasts from REAL(8) to COMPLEX(8), which were apparently OK in older versions. One could probably fiddle around with the compiler flags to make this work, but that might be worse than the current solution.
- ExaTensor uses `mpifort` to get the linking correct. I've tried with replacing the `mpifort` calls with `gfortran` (and gfortran coming from `gfortran8` or `gfortran11` but then it doesn't find the MPI libraries. Again, one could probably patch this until it works, but at which price.
- GFortran 9 and 10 actually throw internal compiler errors on those files, I guess that would not be an easy fix.

default.nix
sheepforce: Unfortunately not, as OpenMPI does not build with gfortran11, yet. See above.

default.nix
```diff
@@ -130,12 +130,23 @@ let
...
+      };
+
       dkh = callPackage ./pkgs/apps/dkh { };
 
       dftd3 = callPackage ./pkgs/apps/dft-d3 { };
 
       ergoscf = callPackage ./pkgs/apps/ergoscf { };
 
+      exatensor = callPackage ./pkgs/apps/exatensor rec {
```
sheepforce, line 145: Jup, it works like this
```suggestion
      exatensor = callPackage ./pkgs/apps/exatensor rec {
        mpi = super.mpi.override { gfortran = super.gfortran8; };
      };
```

default.nix
sheepforce: This combination works just fine with minimal overrides.

```nix
      dirac = callPackage ./pkgs/apps/dirac rec {
        exatensor = self.exatensor;
      };
      exatensor = callPackage ./pkgs/apps/exatensor rec {
        mpi = super.mpi.override { gfortran = super.gfortran8; };
      };

```

sheepforce:
> Thanks for the contribution. I left a few comments.
> Right now, I can't build the PR. It fails with:
> 
> ```
> /build/DIRAC-21.0-Source/external/stieltjes/stieltjesmod.f90:67:10:
> 
>        use stieltjes_ql_diag
>           1
> Fatal Error: Can't open module file 'stieltjes_ql_diag.mod' for reading at (1): No such file or directory
> ```
> 
> CC @sheepforce: do you have any comments or ideas how to improve the PR?

With my small tweaks in https://github.com/markuskowa/NixOS-QChem/pull/91#discussion_r681866698 it actually builds. Does this fix it for you?

default.nix
markuskowa: > This combination works just fine with minimal overrides.

That's probably the best solution.

> OpenMPI does not build with gcc11stdenv due to...

That's good to know. This will most likely affect many more packages than just openmpi(?).

> ExaTensor uses mpifort to get the linking correct. I've tried with replacing the mpifort calls with gfortran (and gfortran coming from gfortran8 or gfortran11 but then it doesn't find the MPI libraries.

The default compiler paths are hard coded into the `openmpi` installation. One can override the default compilers via environment variables such as `OMPI_FC`.

pkgs/apps/exatensor/default.nix
FabianGD: Ok, thats a good point.

default.nix
sheepforce: > > This combination works just fine with minimal overrides.
> 
> That's probably the best solution.
> 
> > OpenMPI does not build with gcc11stdenv due to...
> 
> That's good to know. This will most likely affect many more packages than just openmpi(?).
> 
> > ExaTensor uses mpifort to get the linking correct. I've tried with replacing the mpifort calls with gfortran (and gfortran coming from gfortran8 or gfortran11 but then it doesn't find the MPI libraries.
> 
> The default compiler paths are hard coded into the `openmpi` installation. One can override the default compilers via environment variables such as `OMPI_FC`.

Ah that is good to know. I've changed them in the ExaTensor makefile to gfortran directly and this is obviously not a good solution ^^

markuskowa: For reference with the compiler hard coding: https://github.com/NixOS/nixpkgs/blob/a6a026abe97faa42db6ccd48f3be0b8baf4cb26f/pkgs/development/libraries/openmpi/default.nix#L104

(MPICH and MVAPICH follow the same logic). 

You can in principal use gfortran directly if you add the right compiler flags, obtained from `mpifortran -showme:compile` or `mpifortran -showme:link`.


(The disadvantage of both solutions is that the keywords/variable names are specific to openmpi.)

default.nix
markuskowa: Let's try that solution first, then.

markuskowa:
I still get this error, when trying to build it against the latest `nixpkgs-unstable` (4d3e13e51b6e59b7892874b03072a7c931911e66)
```
[ 42%] Building Fortran object src/grid/CMakeFiles/grid.dir/num_grid_cfg.F90.o
warning: failed to parse dependencies from Fortran source '/build/DIRAC-21.0-Source/src/reladc/adccore.F': syntax error, unexpected WORD
[ 42%] Building Fortran object src/grid/CMakeFiles/grid.dir/grtest.F.o
[ 42%] Building Fortran object src/grid/CMakeFiles/grid.dir/Lebedev-Laikov.F.o
[ 42%] Building Fortran object src/grid/CMakeFiles/grid.dir/num_grid_gen.F90.o
[ 43%] Building Fortran object src/reladc/CMakeFiles/reladc.dir/__/__/external/stieltjes/stieltjes_ql_diag.f90.o
[ 43%] Building Fortran object src/reladc/CMakeFiles/reladc.dir/adc_cfg.F90.o
[ 43%] Building Fortran object src/reladc/CMakeFiles/reladc.dir/__/__/external/stieltjes/stieltjesmod.f90.o
[ 43%] Building Fortran object src/reladc/CMakeFiles/reladc.dir/adc_mat.F90.o
[ 43%] Building Fortran object src/reladc/CMakeFiles/reladc.dir/adc_fano_exchange.F90.o
[ 44%] Building Fortran object src/luciarel/CMakeFiles/luciarel.dir/krci_cfg.F90.o
/build/DIRAC-21.0-Source/external/stieltjes/stieltjesmod.f90:67:10:

   67 |       use stieltjes_ql_diag
      |          1
Fatal Error: Cannot open module file 'stieltjes_ql_diag.mod' for reading at (1): No such file or directory
compilation terminated.
```

FabianGD:
> I still get this error, when trying to build it against the latest nixpkgs-unstable (4d3e13e51b6e59b7892874b03072a7c931911e66)

Ok, I'll retry with this specific version and see what is the issue here. Thanks!

pkgs/apps/dirac/default.nix
```diff
@@ -0,0 +1,131 @@
...
+    mpi
+  ];
+
+  src = fetchurl {
+    url = "https://zenodo.org/record/4836496/files/DIRAC-${version}-Source.tar.gz";
+    sha256 = "0vqmin24xqkcp3rvml0dimc4kjmb3m37c3l6bkjy4y6d3xyb41yg";
+  };
+
+  prePatch = ''
+    ls -lah
```
markuskowa, line 41:
```suggestion
```

pkgs/apps/dirac/default.nix
```diff
@@ -0,0 +1,131 @@
...
+
+  patches = [
+    ./0001-exatensor-cmake.patch
+  ];
+
+  postPatch = ''
+    substituteInPlace ./cmake/custom/exatensor.cmake \
+    --subst-var-by exatensor ${exatensor}
+
+    patchShebangs .
```
markuskowa, line 52: That will go through all script and sources. Can we narrow that down too specific scripts or subdirs?

pkgs/apps/dirac/default.nix
```diff
@@ -0,0 +1,131 @@
...
+  '';
+
+  FC = "mpif90";
+  CC = "mpicc";
+  CXX = "mpicxx";
```
markuskowa, line 57: Does the project use C++?

pkgs/apps/dirac/default.nix
```diff
@@ -0,0 +1,131 @@
...
+  weird hack in the install.cmake to install pam into the bin folder.
+
+  The pam script is just a Python script that sets a ton of ENV variables.
+  */
+  postFixup = ''
+    mv $out/bin/pam-dirac $out/bin/pam
+
+    substituteInPlace $out/bin/pam \
+      --replace "find_executable('mpirun')" "'${mpi}/bin/mpirun'"
+
```
markuskowa, line 91:
```suggestion
```

pkgs/apps/exatensor/default.nix
```diff
@@ -0,0 +1,54 @@
...
+    if (blas.passthru.implementation == "openblas") then "OPENBLAS"
+    else if (blas.passthru.implementation == "mkl") then "MKL"
+    else throw "Only MKL and OPENBLAS supported by ${pname}.";
+
+  PATH_BLAS_OPENBLAS = blas.passthru.provider;
+  PATH_BLAS_MKL = blas.passthru.provider;
+  PATH_BLAS_MKL_DEP = "${blas.passthru.provider}/lib";
+  PATH_BLAS_MKL_INC = "${blas.passthru.provider}/include";
+
+  installPhase = ''
```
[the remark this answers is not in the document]
markuskowa: I think there are 2 spaces indent.

pkgs/apps/dirac/default.nix
```diff
@@ -0,0 +1,131 @@
...
+  /*
+  Make the MPI stuff available to the DIRAC script by hard-coding the MPI path.
+  Moving pam-dirac to pam to conform with the DIRAC manual. Usually there is a
+  weird hack in the install.cmake to install pam into the bin folder.
+
+  The pam script is just a Python script that sets a ton of ENV variables.
+  */
+  postFixup = ''
+    mv $out/bin/pam-dirac $out/bin/pam
```
markuskowa, line 87: Should the name of the target not be `pam-dirac`? Is there a reason it gets modified/shortened?

FabianGD: Yes, the documentation of DIRAC always refers to the script as `pam`. Internally in the CMake process it copies the file called `pam` to `pam-dirac` and then does some weird trick to hack CMake installation. But in the end, it wants to have a main entrypoint called `pam`.

FabianGD:
The DIRAC derivation should be building now. However, I needed to remove the enableParallelBuilding attribute, as otherwise we would get bugs like you @markuskowa got while compilation (`use stieltjes_ql_diag`). I got a similar bug (sometimes) myself for a different subpackage (`use adc_fano_diag`).

pkgs/apps/dirac/default.nix
FabianGD: Yes, for some modules like in the `src/xcfun` package.

pkgs/apps/dirac/default.nix
FabianGD: Those Python scripts are unfortunately scattered throughout the project (it's a real mess), so, no, I don't think so. :)

pkgs/apps/dirac/default.nix
sheepforce: XCFun is part of upstream nixpkgs. Can you use this one?

pkgs/apps/dirac/default.nix
sheepforce: If you just want to patch python sources, you could use find.
```
find -type f -name "*.py" -exec patchShebangs {} \;
```

pkgs/apps/dirac/default.nix
markuskowa: @sheepforce The sources are shipped with Dirac (no git pull etc.). I don't think we need to replace them with upstream sources (that may just cause unnecessary trouble).

sheepforce: I see. I thought those were just API wrappers for XCFun. :slightly_smiling_face:

markuskowa:
@FabianGD Looks good to me now. I can build it now. Last thing before we merge: can you please squash the commits into 2 commits. One for exatensor, one for dirac, in the standard message format (e.g. exatensor: init at 2020-07-15).

pkgs/apps/dirac/default.nix
markuskowa: That is a good solution. Please implement that.

FabianGD: However, this will not work, as a significant portion of the Python files do not have an ending (or at least not .py). 🙈  Furthermore, there is a number of .sh files that are also fixed in the process.

sheepforce: OK, that is annoying. We could improve the find approach a little bit then:
```
# Shell scripts by suffix
find -type f -name "*.sh" -exec patchShebangs {} \;

# Python scripts by shebangs
find -type f -exec grep '#!/usr/bin/env python' {} \; -exec patchShebangs {} \;
```

markuskowa: `find -exec patchShebangs` actually does not work (`find: 'patchShebangs': No such file or directory`, I guess because it is defined as function).

The find version with `grep` does not make it faster. I guess in this case we can keep `patchShebangs .` and rely on it to do its job.

FabianGD: IMO, looking at the alternatives, `patchShebangs .` seems the most readable solution for this situation. :)

markuskowa (approved):

PR #92, opened 2021-08-03
Allow current Molcas 21.06 in SHARC 2.1

sheepforce:
We haven't updated the Molcas patch in sharc, which allows to use newer versions. This broke Molcas in SHARC. This should again fix it.


markuskowa:
Good catch.

PR #95, opened 2021-08-11
Q-Chem: init at 5.{1..4}

sheepforce:
This is an attempt to add Q-Chem to the overlay. Q-Chem installation is unfortunately very tricky with Nix and I could use some pointers. Q-Chem puts a major obstacle in my way. The license/registration process:

Q-QChem has an interactive installation script, that I attempt to answer in the `installPhase`. This installation script wants to either have network access, to request a license-file on the fly from the Q-Chem servers, or asks for an order number and registered mail address (better). It then figures out, the machines name and hardware (CPU cores and GPUs mainly) and writes a file, that is to be submitted to the Q-Chem mail address manually, which in turn will send a license token file, registering each node on which Q-Chem shall run. This must be put in the Q-Chem installation directory. I have absolutely no good idea how to handle this. The only thing I could come up with is to make a second derivation with the license token in a separate store path and then `symlinkJoin` them later. But this is a very manual approach and I have no idea how to get this reproducible. The license file and its hash changes, of course.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ The latest builds for the master branch and stable version are stored on [Cachix
 * Cache URL: https://nix-qchem.cachix.org
 * Public key: nix-qchem.cachix.org-1:ZjRh1PosWRj7qf3eukj4IxjhyXx6ZwJbXvvFk3o3Eos=
 
-If you are allowed to add binary substituters (as trusted user), 
+If you are allowed to add binary substituters (as trusted user),
 you may simply add it with `nix-shell -p cachix --run "cachix use nix-qchem"`.
 
 ## Configuration
@@ -47,6 +47,26 @@ you may simply add it with `nix-shell -p cachix --run "cachix use nix-qchem"`.
 The overlay can be configured either via an attribute set or via environment variables.
 If no attribute set is given the configuration the environment variables are automatically
 considered (impure).
+
+### Special Installation Instructions
+
+#### Q-Chem
+The Q-Chem version `5.{1..4}` are packaged. Download the Linux binaries with all options enabled for your respective version at [https://www.q-chem.com/install/#linux](https://www.q-chem.com/install/#linux).
+Q-Chem is evaluated in two steps to obtain a valid license, after the installer has run.
+
+  1. Build the installer `nix-build -A qchem.q-chem-installer`.
+     This will install Q-Chem into the store and prepare a preliminary `license.data` file, and prepare a script, that helps you to obtain the final `license.data`.
+     The `qchem.q-chem-installer.getLicense` attribute (available as `./result/bin/q-chem_prep_license`) requires the following environment variables
+
+     - `$QCHEM_NODES`: a space-separated list of nodes, for which a Q-Chem license should be obtained. All nodes must be reachable via MPI.
+     - `$QCHEM_MAIL`: the e-mail address associated with the Q-Chem license. The license file will be sent to this address by Q-Chem.
+     - `$QCHEM_ORDNUM:` the order number for Q-Chem.
+
+     After these variables have been set, run `./result/bin/q-chem_prep_license`. You should now have `./license.data`. Send this file via mail to `license@q-chem.com`.
+
+  2. After you have received your license file from `license@q-chem.com`, point `$NIXQC_LICQCHEM` or `licQChem` to this file.
+     The `qchem.q-chem` attribute can be used normally, now; i.e. `nix-build -A qchem.q-chem`.
+
 
 ### Configuration via nixpkgs
 Configuration options can be set directly via `config.qchem-config` alongside other nixpkgs config options.
@@ -56,8 +76,9 @@ Configuration options can be set directly via `config.qchem-config` alongside ot
 * `srcurl`: URL for non-free packages. If set this will override the `requireFile` function of nixpkgs to pull all non-free packages from the specified URL
 * `optpath`: Path to packages that reside outside the nix store. This is mainly relevant for Gaussian and Matlab.
 * `licMolpro`: Molpro license token string required to run molpro.
-* `optArch`: Set gcc compiler flags (mtune and march) to optmize for a specfic architecture. Some upstream packages will be overriden to use make use of AVX (see `nixpkgs-opt.nix`). Note, that this also overrides the stdenv
+* `optArch`: Set gcc compiler flags (`mtune` and `march`) to optimize for a specific architecture. Some upstream packages will be overriden to use make use of AVX (see `nixpkgs-opt.nix`). Note, that this also overrides the stdenv
 * `useCuda`: Uses Cuda features in selected packages.
+* `licQChem`: Path to a Q-Chem license file as obtained via mail.
 
 
 ### Configuation via environment variables
@@ -70,3 +91,4 @@ The overlay will check for environment variables to configure some features:
 * `NIXQC_AVX`: see `optAVX`, setting this to 1 corresponds to `true`.
 * `NIXQC_OPTARCH`
 * `NIXQC_CUDA`: see `useCuda`, setting this to 1 corresponds to `true`.
+* `NIXQC_LICQCHEM`: see `licQChem`
--- a/overlay.nix
+++ b/overlay.nix
@@ -191,6 +191,8 @@ let
 
         pysisyphus = super.python3.pkgs.toPythonApplication self.python3.pkgs.pysisyphus;
 
+        q-chem-installer = callPackage ./pkgs/apps/q-chem/installer.nix { };
+
         qdng = callPackage ./pkgs/apps/qdng {
           stdenv = aggressiveStdenv;
           protobuf = super.protobuf3_11;
@@ -337,6 +339,8 @@ let
         molpro20 = null;
         molpro-ext = null;
 
+        q-chem = null;
+
         # Provide null gaussian attrs in case optpath is not set
         gaussian = null;
       } // lib.optionalAttrs (cfg.licMolpro != null) {
@@ -354,6 +358,10 @@ let
 
         molpro-ext = callPackage ./pkgs/apps/molpro/custom.nix { token = cfg.licMolpro; };
 
+      } // lib.optionalAttrs (cfg.licQChem != null) {
+        q-chem = callPackage ./pkgs/apps/q-chem/default.nix {
+          qchemLicensePath = cfg.licQChem;
+        };
       } // lib.optionalAttrs (cfg.optpath != null) {
         #
         # Quirky packages that need to reside outside the nix store
--- a/pkgs/apps/q-chem/installer.nix
+++ b/pkgs/apps/q-chem/installer.nix
@@ -1,0 +1,132 @@
+{ stdenv, lib, requireFile, autoPatchelfHook, tcsh, perl, openssh
+, writeScript, runtimeShell, hostname
+, minorVersion ? 1
+} :
+
+assert
+  lib.asserts.assertMsg
+  (builtins.elem minorVersion [ 1 2 3 4 ])
+  "Unsupported version of Q-Chem";
+
+let
+  url = "https://www.q-chem.com/install/#linux";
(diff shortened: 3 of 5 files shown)


aolasz:
The license file is the same for all machines. The only catch is that when you request new license for other machines, the license file gets appended; and only the appended license file will work on the other machines.

markuskowa:
> The license file is the same for all machines. The only catch is that when you request new license for other machines, the license file gets appended; and only the appended license file will work on the other machines.

That is not too bad. I guess it will not happen too frequently that you need to add new machines. Providing one license file, like we do for molpro, seems to be an acceptable solution?

aolasz:
The molpro way should be fine.

aolasz:
This is how we do it in an overlay in a very manual way (may or may not be helpful):  

```nix
#overlay/qchem-5.4/default.nix
{ stdenvNoCC, requireFile, buildFHSUserEnv, writeScript, runtimeShell }:
let
  version = "5.4";

  qchem-unwrapped = requireFile rec {
    name = "qchem${version}";
    # The hash is calculated using nix-hash --type sha256 /nix/store/na0bpz6z3gp4ma7pmhpnqcq7df3ji94s-qchem5.4 after importing into the store
    # Use: pin-path qchem-5.4 /nix/store/na0bpz6z3gp4ma7pmhpnqcq7df3ji94s-qchem5.4
    # to prevent garbage collection on qchem
    sha256 = "39f0fd8500e758e3db5d9ab7e477d2a14232555a17d95f4502e98f4e57b8909a";
    hashMode = "recursive";
    message = ''
      Please prefetch qchem using

      nix-store --add-fixed sha256 --recursive /path/to/${name}
    '';
  };

in buildFHSUserEnv rec {
  name = "qchem-${version}";

  targetPkgs = _: [ qchem-unwrapped ];

  #extraInstallCommands = ''
  #  mv $out/bin/${name} $out/bin/qchem
  #'';

  runScript = writeScript "qchem-run" ''
    #!${runtimeShell}

    export QC=${qchem-unwrapped}
    export QCSCRATCH="''${QCSCRATCH:-/scratch}"
    # exec get_hostid
    exec -- qchem "$@"
  '';
}
```

```nix
#overlay/default.nix
(self: super: let
  pkgs = import super.pkgs {};
  # other stuff
in {
  qchem54 = super.callPackage ./qchem-5.4 { };
})
```

```nix
#default.nix
{ pkgs, ... }: {
  nixpkgs = {
    config.allowUnfree = true;
    overlays = [ (import overlay/default.nix) ];
  };
  environment.systemPackages = with pkgs; [
  qchem54
  ];
}
```
First the line `exec -- qchem "$@"` gets commented out and `exec get_hostid` gets uncommented so when we run the command `qchem-5.4` it runs `get_hostid`. The host ids get collected and sent for the license. Once we get the license file we copy it to `qchem5.4/qcaux/license/`. We comment out `exec get_hostid`, and uncomment `exec -- qchem "$@"`. Then we update the store and redeploy the changed qchem.  
There must be a better way.

sheepforce:
Thank your very much for the input, @aolasz :slightly_smiling_face: I will be on vacation for a few days now, but take some inspiration from your approach when I come back.

sheepforce:
Alright, after a too long time I've finally got it somehow running. Q-Chem evaluates to three derivations in two stages:

1. `nix-build -A qchem.q-chem.installer`
    1. runs the installer and sets up the directory structure. patchelf runs here to get all binaries right and interpreter paths are patched. Currently I've only hardcoded the SMP parallel version. The installer is not able to retrieve license data, due to missing network connection in the sandbox.
    2. builds a script, that must be executed impurely, and which dials into a list of nodes to be activated via MPI, to obtain license data for them. The script builds license.dat, which can then be sent via mail to license@q-chem.com.
2. `nix-build -A qchem.q-chem.program` After the license file has been obtained, `licenseQChem` must point to the file with the license. Symlink and wrapper magic comes into play, that finalises the Q-Chem installation, making it actually usable, without polluting the store with Q-Chem copies.

I could not solve one problem, though; , `ordQChem`, `mailQChem` and `licenseQChem` are always `null`, even when I set them via `qchem-config` in the package import. Thus, there are still trace and debug statements left. @markuskowa can you maybe help?

markuskowa:
> I could not solve one problem, though; , ordQChem, mailQChem and licenseQChem are always null, even when I set them via qchem-config in the package import. Thus, there are still trace and debug statements left.

Can you post a code snipet that shows how you use `qchem-config`?

markuskowa:
The basic setup looks okay to too me. A few suggestions:
* obtaining the license looks like a step that should be carried manually and only once (if I understand the procedure correctly). Thus I am not sure how meaningful it is to put `ordQChem` and `mailQChem` into the config. It could be just a command line option for the script. This is also confidential information that allows you to administer the license and as such you probably do not want that publicly visible in the system.
* `licenseQChem` -> `licQChem`, to follow the same scheme we already have for molpro.
* I do not have a QChem license to test this. I will rely on you make sure that it works.
* Rename the two attributes to `q-chem` and `q-chem-installer`, seems more intuitive.

sheepforce:
> if I understand the procedure correctly

^^ Relatable. It's definitely different from the other derivations then. I would add a few words in the README, so that the procedure becomes clear.

> * It could be just a command line option for the script. This is also confidential information that allows you to administer the license and as such you probably do not want that publicly visible in the system.

Hm, you are right, but I am not sure how to handle this. The setup script of q-chem already requires this information and checks if those are valid strings. As the script refuses to continue the installation if no valid values are passed, I have to supply *something*. I may try with placeholders and substitute them in the bash script I am writing. Maybe this works.

aolasz:
I am very excited to see progress. I would gladly try out your PR.  I can test for Q-Chem versions 5.4.2, 5.3.2, and 5.1.0.
  
I need instructions, though, as even the intended way of using NixOS-QChem is not clear to me (I'm quite new to the whole nix concept).

sheepforce:
@markuskowa ready for review, I am happy now. :slightly_smiling_face: 

@aolasz I've updated the expression and the versions 5.1 to 5.4 should now work in the SMP version. We haven't found good use for the MPI parallel versions yet, but I can later try to also build the MPI-versions, if those would be helpful to you. I've also added some instructions for Q-Chem in the README in this branch. Regarding the usage of NixOS-QChem I can provide you a small snippet of code to directly execute (bottom of this comment). You can also have a look at our recently published paper https://doi.org/10.1002/qua.26872 for some use cases, the [nixpkgs chapter on overlays](https://nixos.org/manual/nixpkgs/stable/#chap-overlays) in general, and some of our examples, e.g. [a complete example of a pinned shell for a coomputational chemistry project](https://github.com/markuskowa/NixOS-QChem/blob/fc118b5e5bfd74a1bc9c3784377974c09d36ba36/examples/pinned-project-shell/shell.nix). If you need more input, just tell me. :slightly_smiling_face: 

Here is an unholy "one"-liner to test Q-Chem (assuming you already have your qchem.license.dat file somewhere) :laughing: 
```bash
# Add the Q-Chem archive to the store
nix-prefetch-url --type sha256 file:///path/to/qc5x.tar

# Directly launch a nix-shell without any manual steps
nix-shell -E '
let
  qchemOvl = import (builtins.fetchTarball "https://github.com/sheepforce/NixOS-QChem/archive/refs/heads/qchem.zip");
  pkgs = import (builtins.fetchTarball "https://github.com/NixOS/nixpkgs/archive/5efc8ca954272c4376ac929f4c5ffefcc20551d5.tar.gz") {
    overlays = [ qchemOvl ];
    config = {
      allowUnfree = true;
      licQChem = "/path/to/your/qchem.license.dat"; 
    };
  };
in with pkgs; mkShell {
  buildInputs = [ qchem.q-chem.override { minorVersion = 1; } ];
}
'
```

README.md
```diff
@@ -39,9 +39,29 @@ The latest builds for the master branch and stable version are stored on [Cachix
 * Cache URL: https://nix-qchem.cachix.org
 * Public key: nix-qchem.cachix.org-1:ZjRh1PosWRj7qf3eukj4IxjhyXx6ZwJbXvvFk3o3Eos=
 
-If you are allowed to add binary substituters (as trusted user), 
+If you are allowed to add binary substituters (as trusted user),
 you may simply add it with `nix-shell -p cachix --run "cachix use nix-qchem"`.
 
+### Special Installation Instructions
```
markuskowa, line 45: I would move that section below configuration.

README.md
```diff
@@ -70,3 +91,4 @@ The overlay will check for environment variables to configure some features:
 * `NIXQC_AVX`: see `optAVX`, setting this to 1 corresponds to `true`.
 * `NIXQC_OPTARCH`
 * `NIXQC_CUDA`: see `useCuda`, setting this to 1 corresponds to `true`.
+* `NIXQC_LICQCHEM`: see `licQChem`, setting this to 1 corresponds to `true`.
```
markuskowa, line 94: This is not a bool?

overlay.nix
```diff
@@ -191,6 +191,11 @@ let
 
         pysisyphus = super.python3.pkgs.toPythonApplication self.python3.pkgs.pysisyphus;
 
+        q-chem-installer = callPackage ./pkgs/apps/q-chem/installer.nix { };
+        q-chem = callPackage ./pkgs/apps/q-chem/default.nix {
```
markuskowa, line 195: Please put `q-chem` after the molpro section and make it conditional (e.g. `// lib.optionalAttrs (cfg.licQChem != null)`). That will avoid evaluation errors, when there is no license present.

README.md
sheepforce: Completely right, copy paste error.

pkgs/apps/q-chem/installer.nix
```diff
@@ -0,0 +1,132 @@
...
+    '';
+
+    dontConfigure = true;
+    dontBuild = true;
+
+    # Run the interactive configuration script.
+    # It will prepare a file to be sent to license@q-chem.com,
+    # but the HostIds will be missing, as get_hostid has not been fixed yet.
+    installPhase = ''
+      ./qcinstall.sh << EOF
```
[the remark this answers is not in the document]
sheepforce: I am not entirely sure how exactly your alternative idea would look like. Basically replacing/patching `qcinstall` with our own nixified script, execute it interactively, imperatively, and with network connection and installing somewhere, e.g. `/opt/q-chem` and then wrapping as with Gaussian?

Certainly, the installer complicates things and the symlink stuff to avoid a second copy in the store is also *interesting*. But it has some advantages over the above idea (where I am not certain that you mean exactly this);

  - it is functional and involves no imperative step, that somehow influences anything besides the license.
  - If I should add the MPI versions later, the derivation is aware of the MPI requirement and make the appropriate settings from the context of the derivation
  - external installations in `/opt` or whatever require the sandbox path settings and those are only accessible for an admin. Thus, the Q-Chem installation would be exclusive to admins and normal users could not use the Q-Chem derivation.

The last point would be crucial for us, but as I said; maybe I got you wrong.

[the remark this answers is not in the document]
sheepforce: Also, I don't understand `qcinstall.sh` completely. It's a 750 lines bash script, which then again also does not everything, as far as I understand it. The complete installation procedure creates more files and directories, than what I see in the script.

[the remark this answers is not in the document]
sheepforce: @markuskowa as by your suggestions I also rewrote the activation script to use plain ssh instead of MPI. However, there is a problem, that we didn't think about. If of course requires and SSH server running. In my case, where I just want to activate my local workstation and no SSH server is running, simply dialing in via SSH doesn't work of course. MPI then again, works, as it figures out that for localhost SSH is not required. I tried to circumvent it now with a small bash if.

markuskowa (approved):
Looks good to me now. I would suggest we merge it as is now and see how it works in practice.

sheepforce:
I agree :slightly_smiling_face: @aolasz if you find anything odd or are interested in the MPI versions just open an issue and I will see if I can fix it.

PR #97, opened 2021-09-02
Allow for gcc optmization flags

markuskowa:
This PR cleans up the optimization of upstream packages and adds the possibility of adding gcc optimization flags to stdenv.

Up till now the overlay applied overrides upstream packages directly in the nixpkgs name space. This is in fact a bug as it causes mass rebuilds on packages that not related to the overlay (e.g. the fftw overrides causes a rebuild of firefox). This has been changed. The overrides only affect packages in the `qchem` subset. 

A new config option `optArch` is introduced, which adds `march` and `mtune` flags to the stdenv.

### DIFF
--- a/pkgs/apps/bagel/default.nix
+++ b/pkgs/apps/bagel/default.nix
@@ -2,7 +2,6 @@
 , makeWrapper, openssh
 , python3, boost, mpi, blas, lapack
 , scalapack ? null
-, optAVX ? true
 } :
 
 let
@@ -44,7 +43,6 @@ in stdenv.mkDerivation rec {
     "-O3"
     "-DCOMPILE_J_ORB"
   ] ++ lib.lists.optionals (!useMKL) [ "-lblas" "-llapack" ]
-    ++ lib.lists.optional optAVX "-mavx -mavx2"
     ++ lib.lists.optional (blas.passthru.implementation == "openblas") "-DZDOT_RETURN"
   );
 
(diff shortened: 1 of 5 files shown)


pkgs/apps/bagel/default.nix
```diff
@@ -44,7 +43,6 @@ in stdenv.mkDerivation rec {
     "-O3"
     "-DCOMPILE_J_ORB"
   ] ++ lib.lists.optionals (!useMKL) [ "-lblas" "-llapack" ]
-    ++ lib.lists.optional optAVX "-mavx -mavx2"
```
sheepforce, line 47: Is this supposed to be handled by the `march` optimisation from the `stdenv`? Wouldn't it be better to keep using the `stdenv.hostPlatform.avx2Support` flags directly? The impurity of `march` and `mtune` doesn't feel so nice.

markuskowa: Yes, the idea is that the march/mtune flags are set in the new stdenv. This way more of the code is compiled with optimizations and we need to mention it in less places in derivations (unless of course it is a configure or cmake flag).
Note that the march/mtune flags are perfectly pure here, when set to a specific architecture. These optimizations only become impure when set to `-march=native`/`-mtune=native`.

sheepforce: Ah I see, this makes a lot of sense. Thanks for the explanation.

PR #98, opened 2021-09-06
Psi4: 1.3.2 -> 1.4, libint1: removed

sheepforce:
Update to the recent release of Psi4. The CMake build system is still super hacky and required new workarounds for libint. Libint for Psi4 requires that many customisations, that I don't see how we could fit it into the upstream derivation of libint. Instead I am using a minor CMake patch to force building Psi4's custom Libint version together with the rest of Psi4. On the upside, I was able to remove libint1, as Psi4 was the last package still using it. The downside is, that the Psi4 build now takes half an eternity, as it compiles libint with large angular momentum.


sheepforce:
See https://github.com/psi4/psi4/issues/1302 regarding upstream Libint in Psi4.

PR #101, opened 2021-09-10
pysisyphus: 0.7.post1 -> 0.7.2

sheepforce:
Update of Pysisphus with important updates to TRIC coordinate optimisation and excited state optimisation.

Requires https://github.com/NixOS/nixpkgs/pull/137267 to appear in nixpkgs-unstable first.


sheepforce:
Alternatively we remove the strict version constraint with a patch. @eljost this minor release is not a problem, correct?

PR #103, opened 2021-09-14
XTBPATH Environment

sheepforce:
This always sets the `XTBPATH` environment, not just for the executable. This is relevant for the C- and Python-APIs, which do not see this variable otherwise and can't perform GFN-0 calculations.

cc @eljost

### DIFF
--- a/pkgs/apps/xtb/default.nix
+++ b/pkgs/apps/xtb/default.nix
@@ -50,9 +50,10 @@ in stdenv.mkDerivation rec {
 
   postFixup = ''
     wrapProgram $out/bin/xtb \
-      --prefix PATH : "${binSearchPath}" \
-      --set-default XTBPATH "$out/share/xtb"
+      --prefix PATH : "${binSearchPath}"
   '';
+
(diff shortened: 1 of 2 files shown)


pkgs/apps/xtb/default.nix
```diff
@@ -50,10 +50,11 @@ in stdenv.mkDerivation rec {
 
   postFixup = ''
     wrapProgram $out/bin/xtb \
-      --prefix PATH : "${binSearchPath}" \
-      --set-default XTBPATH "$out/share/xtb"
+      --prefix PATH : "${binSearchPath}"
   '';
 
+  setupHooks = [ ./xtbHook.sh ];
```
markuskowa, line 56: This is for python + nix-shell?

sheepforce: Yes and no. It also is important for Pysisyphus GFN-0 (and other cases like this) calculations by XTB-Python. The C-API and Python API will also not work for GFN-0, if this variable is not set, even with a regular `nix-build`. The old version with the wrapper just around the executable does not work, as the executable itself will never be called.

PR #107, opened 2021-09-29
Installer script

markuskowa:
Draft for the installer.
Setup tests with 
* [x] Arch Linux
* [x] Ubuntu
* [x] Debian
* [ ] Fedora

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -3,7 +3,7 @@
 set -e
 
 error() {
-  printf "$0: $@" >&2
+  printf "$0: $@\n" >&2
   exit 1
 }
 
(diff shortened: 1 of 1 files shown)


sheepforce:
Works on Debian Bullseye.

install.sh
```diff
@@ -0,0 +1,132 @@
...
+EOF
+else
+  printf "Config file $config_file exists. Skipping config creation."
+fi
+
+
+printf "\n\n"
+printf "The Nix and the NixOS-QChem overlay are now\n"
+printf "installed in you system. Before you can use it\n"
+printf "you need refresh your by launching a fresh shell or doing re-login'\n"
```
sheepforce, line 126: New shell and sourcing the bashrc is not enough at least in Debian and Ubuntu.

sheepforce (changes requested):

sheepforce:
Works on Ubuntu 20.04 LTS.

install.sh
```diff
@@ -0,0 +1,132 @@
...
+EOF
+
+  exit 0
+fi
+
+# Check if git is available
+command -v git --version > /dev/null || error "Git is not installed. We need git to check out the overlay."
+
+# Needed by the nix installer
+command -v curl --version > /dev/null || error "Curl is not installed."
```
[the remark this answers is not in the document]
markuskowa: I added a new line to the error message.

install.sh
markuskowa: Yes, the issue is that the PATH is updated via `/etc/profile` and not `bashrc`. What you need is a fresh login shell.

sheepforce:
Fedora does NOT work with a minor linking error in the user environment
[nix-install.log](https://github.com/markuskowa/NixOS-QChem/files/7286410/nix-install.log)
:

markuskowa:
> Fedora does NOT work with a minor linking error in the user environment [nix-install.log](https://github.com/markuskowa/NixOS-QChem/files/7286410/nix-install.log) :

Bummer, but not our fault as it is the nix installer quitting (not clear to me why this happens).

PR #109, opened 2021-10-04
turbomole: fixing patchelf segfaults

### LINKED ISSUE #108: Turbomole Broken with PatchELF
Some recent Nixpkgs commit broke Turbomole. Now, patchelf throws multiple segfaults and the derivation crashes. I am a little bit lost. The last update to patchelf happened in August and does not seem to be the cause of the problem. @markuskowa Do you have any idea if something important has changed in patchelf or stdenv, that breaks this build? I have attached the build log.
[turbomole.log](https://github.com/markuskowa/NixOS-QChem/files/7277211/turbomole.log)

sheepforce:
Closes #108 

I am not entirely sure what is going on there, but this fixes it. Although everything except MPI is statically linked in turbomole, the executables still set an interpreter for shared libraries. Setting the interpreter is necessary for all executables. But attempting to add an rpath for executables only works for the MPI versions.

### DIFF
--- a/pkgs/apps/turbomole/default.nix
+++ b/pkgs/apps/turbomole/default.nix
@@ -97,6 +97,7 @@ in stdenv.mkDerivation rec {
   '';
 
   dontAutoPatchelf = true;
+  noAuditTmpdir = true; # Patchelf segfaults on multiple files and crashes the build
 
   postFixup = let
     libSearchPath = lib.strings.makeSearchPath "lib" [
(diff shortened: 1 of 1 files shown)


pkgs/apps/turbomole/default.nix
```diff
@@ -107,15 +108,19 @@ in stdenv.mkDerivation rec {
...
-    # Patch elf dynamic library paths.
+    # Patch elf dynamic library paths and interpreters
     for dir in $TURBODIR/libso $TURBODIR/bin/*; do
-      files=$(find $dir -type f -executable -maxdepth 1 ! -name rimp2 ! -name ccsdf12)
+      files=$(find $dir -type f -executable -maxdepth 1)
       for f in $files; do
-        # echo "Patching $f now!"
-        patchelf \
-          --set-interpreter "$(cat $NIX_CC/nix-support/dynamic-linker)" \
-          --set-rpath "$libSearchPath" \
```
markuskowa, line 117: The `$libSearchPath` variable is empty here (needs to be `${libSearchPath}`)? patchelf has a segfault bug, when an empty $rpath is used (https://github.com/NixOS/patchelf/pull/319). Could that be it?

markuskowa:
> Although everything except MPI is statically linked in turbomole, the executables still set an interpreter for shared libraries

That means that the binaries are not fully statically linked. Molpro did a similar thing, going back and forth between static, not fully static. You may need to re-evaluate that with every update.

pkgs/apps/turbomole/default.nix
sheepforce: Ouh, you are completely right. This was the issue and is the better solution. Fixed it and works like this.

PR #114, opened 2021-10-21
luscus: init at 0.8.6

markuskowa:
@sheepforce do you have any experience with Luscus? I can build with the expression below and even start the GUI. However, when I open a file the molecule window remains blank.


sheepforce:
Ah another thing: it reads the `*.lus` files generated by `&GIRD_IT` correctly and works as expected. The behaviour you've described seems to occur only with molden files.

Issue #116, opened 2021-10-21
pin nixpkgs version for the overlay


markuskowa:
Right now the overlay is still a bare overlay, which is applied to the system installed `nixpkgs` set. However, this is problematic since it requires a matching version of `nixpkgs`. Locally, I do this by carefully hand selecting nixpkgs/NixOS-QChem versions from my Hydra builds. This is not very user-friendly and potentially error-prone.

A better version could be:
* move the contents of `default.nix` into `overlay.nix`
* provide a `default.nix` which still behaves like an overlay but creates a `qchem` subset with a pinned `nixkgs`.

This would allow for a sensible default, which is guaranteed to work but does not allow for further modifications via additional overlays. If one needs a more customized set, e.g. replace MPI, then one still use the overlay directly via `overlay.nix`

PR #117, opened 2021-10-21
pin nixpkgs in default version of the overlay

markuskowa:
Possible solution to https://github.com/markuskowa/NixOS-QChem/issues/116


markuskowa:
I am going ahead merge this now. It may be that there are some unintended side effects. Please open an issue if you noticed that something broke.

PR #118, opened 2021-10-29
qutip: init at 4.6.2

FabianGD:
This PR packages QuTiP, a quantum computing and simulation library for Python. A colleague of mine needs it for her research. 

I was not entirely sure whether its more sensible to have it here or directly in nixos/nixpkgs, what do you guys think?


FabianGD:
> This looks like straight forward Python package, that could be of interest for a broader audience. I would recommend to open up a PR directly with upstream nixpkgs.

Ok, I'll try there. Thanks! :) @sheepforce Any other opinions?

FabianGD:
For future reference, my attempt to getting it into nixpkgs is found here: https://github.com/NixOS/nixpkgs/pull/144102

PR #122, opened 2021-11-30
psi4: 1.4.1 -> 1.5

sheepforce:
Update to the latest upsteam release, unfortunately still with the original Libint2 hack as upstream (the build times are terrible).
https://github.com/psi4/psi4/releases/tag/v1.5

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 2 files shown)


sheepforce:
A flake update is required to get qcelemental 0.24.0, which is a dependency of Psi4 and has been merged recently into nixpkgs-unstable.

PR #123, opened 2021-12-07
molpro: 2021.2.1 -> 2021.3.1

markuskowa:


markuskowa:
OK, it looks like there is an MPI related bug in 2021.3.1. It fails at various steps in the job (optg, MRCI).

PR #126, opened 2021-12-14
Pysisyphus: 0.7.2 -> 0.7.4

sheepforce:
An update of the Pysisyphus program. I've also changed how optional dependencies are passed and follow the upstream approach with the enableX pattern, instead of passing null packages. Also switched to pytestCheckHook instead of explicit pytest calls and overall everything should be more canonical now.
XTB and Pysisyphus are included in their completely free versions (build with `allowUnfree = false` now).


pkgs/apps/pysisyphus/default.nix
```diff
@@ -79,42 +89,50 @@ in
...
       rev = version;
-      sha256 = "wO/D7ySH0g/qN2aqzOF2Be3aw3U248dvuIEaTAkFYC4=";
+      hash = "sha256-AvJ/9+63yxfeCvL2gRmdZMt/GbvRzMbwlGgOwS2Rrek=";
     };
 
-    patches = [
-      ./scikit-learn.patch
-      ./h5py.patch
-    ];
+    # patches = [
```
markuskowa, line 117: This is not needed anymore?

overlay.nix
```diff
@@ -238,11 +238,7 @@ let
 
         wfoverlap = callPackage ./pkgs/apps/wfoverlap { };
 
-        xtb = callPackage ./pkgs/apps/xtb {
-          turbomole = null;
-          cefine = null;
-          orca = self.orca;
-        };
+        xtb = callPackage ./pkgs/apps/xtb { };
```
markuskowa, line 241: This breaks `qchem.tests.xtb`, since the tests relies on orca.

pkgs/apps/pysisyphus/default.nix
```diff
@@ -1,11 +1,22 @@
...
- autograd, dask, distributed, h5py, jinja2, matplotlib, numpy, natsort, pytest, pyyaml, rmsd, scipy,
- sympy, scikit-learn, qcengine, ase, xtb-python, openbabel-bindings,
+, autograd, dask, distributed, h5py, jinja2, matplotlib, numpy, natsort, pytest, pyyaml, rmsd, scipy
+, sympy, scikit-learn, qcengine, ase, xtb-python, openbabel-bindings, pyscf
  # Runtime dependencies
- runtimeShell, jmol ? null, multiwfn ? null, xtb ? null, openmolcas ? null,
- pyscf ? null, psi4 ? null, wfoverlap ? null, nwchem ? null, orca ? null,
- turbomole ? null, gaussian ? null, gamess-us ? null, cfour ? null, molpro ? null,
+, runtimeShell
+, jmol, enableJmol ? false
```
markuskowa, line 8: All QC packages are now enabled by default. Do you want to a few attributes with your best supported QC packages enabled in the top level file? E.g. `pysisyphus-molcas`, etc.?

markuskowa (changes requested):

pkgs/apps/pysisyphus/default.nix
sheepforce: No they are actually not. All are disabled by default and lazy evaluation of Nix ensures they are never evaluated, unless the corresponding `enableX` attribute is `true` and they are all `false` by default. The `qchem.python3.pysisyphus`, that is available builds with the free quantum chemistry engines by default (XTB, NWChem, Molcas, and JMol and WFoverlap, which are also both free). The other engines are opt-in and the pysisyphus derivation that is exposed is useful for common work and can now be cached by Cachix and public Hydra.

overlay.nix
sheepforce: Fixed

sheepforce:
I've added the `enableX` attributes to the `passthru`, which I now think is in general a good idea. By this, e.g., the XTB test can check if it can use the ORCA tests.

pkgs/apps/pysisyphus/default.nix
markuskowa: Sorry, I meant "disabled by default". I overlooked the `pythonPackages` entries.

PR #136, opened 2022-01-12
updated URL for the  NWChem website

edoapra:
The old NWChem website has been take over by cyber squatters. Please do use the current website https://nwchemgit.github.io


markuskowa:
Thanks for the update!

PR #138, opened 2022-01-19
psi4: external libint + generic blas, dft-d3: replace with free fork

sheepforce:
I've somewhat split up the Psi4 package. Libint now builds as a separate derivation, hopefully drastically improving caching and reducing build times, when dependencies of just Psi4 update. Also I am now allowing Psi4 to build with generic BLAS. The situation in Psi4 improved slightly and threaded BLAS might still be dangerous in some modules, but can always be worked around when setting `OMP_NUM_THREADS=1`. Furthermore I've substituted the DFT-D3 package by the fork of dftb+. It clearly states a GPLv1 license, is written in free form fortran and is otherwise fully compatible to 3.2rev0 of the original program.

In consequence, Psi4 can now be built with `allowUnfree = false`, which is nice, as it allows us to put Psi4 into cachix and the build really takes forever, due to libint.

Still, the custom libint is required until these issues are resolved:
  - https://github.com/psi4/psi4/issues/2398
  - https://github.com/evaleev/libint/issues/190
  - https://github.com/evaleev/libint/pull/205


pkgs/apps/dft-d3/default.nix
```diff
@@ -1,41 +1,29 @@
...
+    hash = "sha256-lda0eEb/QoMG2Sb1/VhJSr+fJcu2wvy1hqw+rVDhe2w=";
   };
 
-  unpackPhase = ''
-    tar -xvf $src
-  '';
-
-  # Using gfortran for linking instead of ifort
-  patches = [
-    ./Linking.patch
```
markuskowa, line 20: The patch can be deleted?

sheepforce: Yes, overlooked this.

pkgs/apps/psi4/default.nix
```diff
@@ -119,18 +148,11 @@ in buildPythonPackage rec {
       sha256 = "sha256-NVpE5fAVWYlkymTrvptZ7xqu68eVy71YJ+dRdBMMU9c=";
     };
 
-    patches = [ ./LibintCmake.patch ];
-
     # Required for Libint compilation. g++ will otherwise not be able to link the large amount of files.
     preConfigure = ''
       ulimit -s 65536
```
markuskowa, line 153: Is that still needed here? It is now in the separate libint calculation.

pkgs/apps/psi4/default.nix
```diff
@@ -36,9 +36,40 @@ let
...
+      "-DENABLE_GENERIC=ON"
+      "-DCMAKE_CXX_FLAGS=-fPIC"
+      "-DENABLE_CXX11API=ON"
+      "-DCMAKECONFIG_INSTALL_DIR=share/cmake/Libint2"
+    ];
+
+    # CMake generates a directory '$out' (not the expanded variable but literally) in the build
+    # directory. Within, the share/cmake files are, and they need to be copied manually now.
+    postInstall = ''
+      cp -r \$out/share $out/.
```
markuskowa, line 71: You can avoid this by using `cmakeFlagsArray` instead of `cmakeFlags`. This should allow for proper shell expansion.

sheepforce: Unfortunately not in this case. It's a bug in this CMake system, not a problem of the shell expansion.

sheepforce: Hm, i will try it nevertheless. Now I am not so sure anymore. :open_mouth:

markuskowa: If  you write `cmakeFlagsArray+=(-DCMAKE_INSTALL_PREFIX=$out)` in the `preConfigure` phase, then you get shell substitution and cmake will never see the literal "$out", but the actual path "/nix/store/...." (?) Am I overlooking something here?

sheepforce: Ah no, probably you are right. I am possibly introducing this indirectly via the `specialInstallCmakeFlags` earlier. I will wait for the build to finish (some hours thanks to libint). If it works, I will push the changes.

PR #139, opened 2022-01-20
wxMacMolPlt: init at 7.7.2

sheepforce:
Adds the wxMacMolPlt prpgram, which is more or less the semi-official GUI for GAMESS-US. Unfortunately I could not figure out how to download the src with fetchurl. Some redirection magic happens there.

### DIFF
--- a/pkgs/apps/wxmacmolplt/default.nix
+++ b/pkgs/apps/wxmacmolplt/default.nix
@@ -1,0 +1,32 @@
+{ stdenv, lib, fetchFromGitHub, wxGTK, libGL, libGLU, pkg-config, xorg, autoreconfHook }:
+
+stdenv.mkDerivation rec {
+  pname = "wxmacmolplt";
+  version = "7.7.2";
+
+  src = fetchFromGitHub {
+    owner = "brettbode";
+    repo = pname;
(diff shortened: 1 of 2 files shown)


pkgs/apps/wxmacmolplt/default.nix
```diff
@@ -0,0 +1,37 @@
...
+  version = "7.7.2";
+
+  /* There is a redirection to another url and it is not directly a file.
+  src = fetchurl {
+    url = "https://uofi.box.com/shared/static/ezm1jd56hlb8s2vsf9ou28vrctjqqa9o.gz";
+    hash = "sha256-7vI8J8ZkYWx0D4gFph2378BV7Y211puvzsaer6/il+Q=";
+  };
+  */
+
+  src = requireFile {
```
markuskowa, line 14: The sources are hosted at Github: https://github.com/brettbode/wxmacmolplt
Why not just fetch either the release tag from there or download the provides source files from there?

sheepforce: I am afraid because I've just overlooked it. Uses Github now. Thanks for the hint :slightly_smiling_face:

PR #140, opened 2022-01-25
gamess-us: compatibility qcengine, pysisyphus: optional gamess

sheepforce:
To work with QCEngine and Pysisyphus, a few tweaks to GAMESS-US are required:
- the socket version is required, thus `enableMpi` allows to disable MPI in GAMESS-US
- the scratch directories need to be set to `./` for qcengine

Pysisyphus was updated to have a GAMESS-US option and now uses the pyproject build instead of deprecated setuptools.

### DIFF
--- a/pkgs/apps/gamess-us/default.nix
+++ b/pkgs/apps/gamess-us/default.nix
@@ -73,8 +73,6 @@ in stdenv.mkDerivation rec {
       export target=${target}
       substituteAllInPlace rungms
 
-      head -n 200 rungms
-
       # Make config accept dynamic OpenBLAS
       substituteInPlace config --replace "libopenblas.a" "libopenblas.so"
       substituteInPlace lked --replace "libopenblas.a" "libopenblas.so"
@@ -123,8 +121,7 @@ in stdenv.mkDerivation rec {
     mkdir -p $out/bin $out/share $out/share/gamess
 
     # Copy the interesting scripts and executables
-    cp gamess.${version}.x rungms $out/bin/.
-    cp $(find -name ddikick.x) $out/bin/.
+    cp gamess.${version}.x rungms ${lib.strings.optionalString (!enableMpi) "ddi/ddikick.x"} $out/bin/.
 
     # Copy the file definitions to share
     cp gms-files.csh $out/share/gamess/.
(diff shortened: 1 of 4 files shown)


pkgs/apps/gamess-us/default.nix
```diff
@@ -1,5 +1,7 @@
 { stdenv, lib, makeWrapper, fetchFromGitLab, requireFile, gfortran, writeTextFile, cmake, perl
 , tcsh, mpi, blas, hostname, openssh, gnused, libxc, ncurses
+, enableMpi ? true
+, scratchHere ? false
```
markuskowa, line 4: Instead of creating this special option for pysisyphus, wouldn't it be simpler if pysisyphus manipulates `$SCRATCH` instead before calling gamess?

sheepforce: You're right. I've changed it. Much better.

pkgs/apps/gamess-us/default.nix
```diff
@@ -121,6 +124,7 @@ stdenv.mkDerivation rec {
 
     # Copy the interesting scripts and executables
     cp gamess.${version}.x rungms $out/bin/.
+    cp $(find -name ddikick.x) $out/bin/.
```
markuskowa, line 127: This seems to fail:
`cp: missing destination file operand after '/nix/store/9pys33yphx6nsqx9qlfxrcwhf1yjx064-gamess-us-2021R2P1/bin/.'`

sheepforce: You are right, the documentation actually explicitly states that ddikick.x is not created if MPI is used ... Should be fixed now.

pkgs/apps/gamess-us/default.nix
```diff
@@ -68,8 +70,11 @@ stdenv.mkDerivation rec {
       # Prepare the rungms script -> replace references to @version@ and @out@
       export mpiname=${mpiname}
       export mpiroot=${mpiroot}
+      export target=${target}
       substituteAllInPlace rungms
 
+      head -n 200 rungms
```
markuskowa, line 76: Does this line have purpose, or was this for debugging purposes?

sheepforce: Unfortunately only debugging. Forgot about it :/

PR #143, opened 2022-01-27
orca: add optional xtb

### LINKED ISSUE #142: calling xtb from orca
Hi,

to be able to use XTB from within ORCA, the 5.0.2 manual states on page 483:

> Only the file `bin/xtb` is used by ORCA. The user should copy this file into the directory where the other ORCA executables are located and rename it as `otool_xtb`.
> **Please use the 6.4.0 version (or any later version) of xtb; older versions are not fully compatible with ORCA anymore; for example, it may not be possible to invoke the solvation model!**

Would it be possible to include this in the qchem.orca package?

Thank you very much
Philipp

sheepforce:
Closes #142 

Enables XTB in ORCA by default.

I've also changed the `requireFile` stuff slightly. The default message seems more helpful to me, as it informs how to add the archive to the nix store, and which of the many ORCA downloads is required.

### DIFF
--- a/pkgs/apps/orca/default.nix
+++ b/pkgs/apps/orca/default.nix
@@ -1,18 +1,15 @@
 { lib, stdenv, requireFile, autoPatchelfHook, makeWrapper
-, openmpi, openssh
+, openmpi, openssh, xtb
 } :
 
-let
-  version = "5.0.2";
-
-in stdenv.mkDerivation {
+stdenv.mkDerivation {
   pname = "orca";
-  inherit version;
+  version = "5.0.2";
 
   src = requireFile {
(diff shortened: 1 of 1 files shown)


pkgs/apps/orca/default.nix
```diff
@@ -1,18 +1,16 @@
 { lib, stdenv, requireFile, autoPatchelfHook, makeWrapper
 , openmpi, openssh
+, xtb, enableXtb ? true
```
markuskowa, line 3: I think there is no need to make this optional. Just add xtb as a dependency (or is there a scenario where one would need to turn it off?).

sheepforce: Probably not. At least I can't imagine any. Is always on now.

PR #145, opened 2022-01-27
readme: adding DOI badge and more explanation on cache and usage

sheepforce:
I thought maybe hinting our paper on this is helpful, as it also explains some motivations. While on it, my coworkers complained that the cachix section is a little bit thin, so I've added a small addition and fixed some typos.


markuskowa:
Nice DOI patch :)

PR #152, opened 2022-01-31
xtb-python: fix meson install

### LINKED ISSUE #151: pysisyphus fails to build with nixpkgs-unstable
pysisyphus started to show build failures, with the latest version of nixpkgs-unstable (e.g. with pinning turned off: `nix-build release.nix --arg pin false -A qchem.pysisyphus`)

```
==================================== ERRORS ====================================
____________ ERROR collecting pysisyphus/tests/test_calculators.py _____________
ImportError while importing test module '/build/source/pysisyphus/tests/test_calculators.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/nix/store/k4rnxn7l5i1jzj33062wlw6g5rlvaydf-python3.9-xtb-python-20.2/lib/python3.9/site-packages/xtb/libxtb.py:32: in <module>
    from ._libxtb import ffi, lib
E   ModuleNotFoundError: No module named 'xtb._libxtb'

During handling of the above exception, another exception occurred:
/nix/store/i6vabb4div9iy6lsl642d86k1q8riasn-python3-3.9.9/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
pysisyphus/tests/test_calculators.py:7: in <module>
    from pysisyphus.calculators import (
pysisyphus/calculators/__init__.py:49: in <module>
    from pysisyphus.calculators.ONIOMv2 import ONIOM
pysisyphus/calculators/ONIOMv2.py:32: in <module>
    from pysisyphus.calculators import (
pysisyphus/calculators/PyXTB.py:3: in <module>
    from xtb.interface import Environment, Param, Calculator as XTBCalculator
/nix/store/k4rnxn7l5i1jzj33062wlw6g5rlvaydf-python3.9-xtb-python-20.2/lib/python3.9/site-packages/xtb/interface.py:24: in <module>
    from .libxtb import (
/nix/store/k4rnxn7l5i1jzj33062wlw6g5rlvaydf-python3.9-xtb-python-20.2/lib/python3.9/site-packages/xtb/libxtb.py:34: in <module>
    raise ImportError("xtb C extension unimportable, cannot use C-API")
E   ImportError: xtb C extension unimportable, cannot use C-API
```

CC @sheepforce

sheepforce:
Closes #151
Caused by the upstream meson update, which now changes how the C-library for XTB is installed. A manual copy fixes this. I've also enabled the XTB-Python tests to track down the problem in the future more easily.

### DIFF
--- a/pkgs/lib/xtb-python/default.nix
+++ b/pkgs/lib/xtb-python/default.nix
@@ -1,5 +1,5 @@
 { buildPythonPackage, lib, fetchFromGitHub, cffi, numpy, ase, qcelemental, meson, ninja, pkg-config,
-  xtb, pytest
+  xtb, pytestCheckHook, python
 }:
 
 buildPythonPackage rec {
@@ -33,13 +33,14 @@ buildPythonPackage rec {
 
   # Build a C module to interface XTB.
   preBuild = ''
-    meson setup build --prefix=$PWD --default-library=shared
+    meson setup build --prefix=$(pwd) --default-library=shared
     ninja -C build install
+    cp ./${python.passthru.sitePackages}/_libxtb.*.so ./xtb/.
   '';
 
-  checkInputs = [ pytest ];
-  checkPhase = "pytest";
-  doCheck = false;
+  checkInputs = [ pytestCheckHook ];
+  pytestFlagsArray = [ "-k 'not qcschema'" ]; # Numerically soooo slightly off
+  pythonImportsCheck = [ "xtb.interface" "xtb.libxtb" ];
 
   meta = with lib; {
     description = "Python wrapper for the semiempirical XTB package";


pkgs/lib/xtb-python/default.nix
```diff
@@ -33,13 +33,14 @@ buildPythonPackage rec {
 
   # Build a C module to interface XTB.
   preBuild = ''
-    meson setup build --prefix=$PWD --default-library=shared
+    meson setup build --prefix=$(pwd) --default-library=shared
     ninja -C build install
+    cp ./${python.passthru.sitePackages}/_libxtb.*.so ./xtb/.
```
markuskowa, line 38: Something is going wrong here:
```
cp: missing destination file operand after './xtb/.'
```

sheepforce: I guess I have an idea. What version does meson report during build? 1.57 or 1.60? This copy works around a changed behaviour of meson with the update in nixpkgs, which changes the installation directory. The original version was OK with 1.57, the version in this branch is OK with 1.60. We can add a conditional, to allow both, but this version fixes #151 at the cost of breaking the current pin.

markuskowa: According to Hydra it is `meson-0.57.1` (from the internally pinned nixpkgs). Should we just stick to the pinned version?

sheepforce: Hm, good question. I would be in favour of updating the pin to have meson 1.60, so that we don't run into this problem later again.
Alternatively, I could replace it with a `find` command, which should work for both meson versions?

markuskowa: Sorry, I got mixed up here. Yes, the purpose is to upgrade the pinning. This version working with meson 1.6 is okay.

markuskowa: (these version bugs are nasty)

PR #154, opened 2022-02-03
gamess-us: provide config answers as commented list for more clarity

sheepforce:


pkgs/apps/gamess-us/default.nix
```diff
@@ -80,36 +80,44 @@ in stdenv.mkDerivation rec {
...
-    no
-    no
-    no
-    no
-    no
-    no
-    no
-    EOF
+  configurePhase =
+    let configAnswers = lib.strings.concatMapStrings (a: a + "\n") ([
```
markuskowa, line 84: More compact?
```suggestion
    let configAnswers = lib.strings.concatStringSep "\n" ([
```

PR #167, opened 2022-02-19
molcas: 21.10 -> 22.02

markuskowa:
libxc is now rquired


sheepforce:
This probably broke the Molcas SHARC interface again, as we need to maintain this patch separately. Will have a look next week if I can automate this.

markuskowa:
@sheepforce which patch do you think broke?

sheepforce:
We had this https://github.com/markuskowa/NixOS-QChem/blob/cb7a02816f917e4fd3295e85cef7115ccb9866bd/pkgs/apps/sharc/molcas_version.patch some time ago, but we removed it apparently. :smile: Nevermind :)

PR #180, opened 2022-04-14
moltemplate: init at 2.20.5

sheepforce:


markuskowa:
> The spell check fails but I would rather not fix spelling in a patch diff ^^

Patches should be filtered out from the check. I need to add that feature.

PR #184, opened 2022-04-26
Cachix Substituter in Flake

sheepforce:
With nix >= 2.6.0 the `nixConfig` became a new configuration attribute in flakes, allowing us to directly set a substituter in the flake. Luckily, this substituter may also be used by unprivileged users, who may normally not be able to a cache.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,6 +3,8 @@
 
   inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
 
+  nixConfig.extra-substituters = [ "https://nix-qchem.cachix.org" ];
+
   outputs = { self, nixpkgs } : let
       lib = import "${nixpkgs}/lib";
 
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -3,6 +3,8 @@
 
   inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
 
+  nixConfig.substituters = [ "https://nix-qchem.cachix.org" "https://cache.nixos.org" ];
```
markuskowa, line 6: Downside: if you have local cache, it will be ignored from here on?

sheepforce: But I have to admit I have no idea how Hydra evaluations would be affected. Possibly hydra decides to use these substituters and ignore your local cache? But as soon as Hydra has evaluated sucessfully, the store of the hydra machine would have those paths anyway. So I would assume that for hydra it is of less importance?

markuskowa: Why do we need `https://cache.nixos.org` in the list then? Here we can safely assume that this is part of the system configuration.

sheepforce: When you accept the cache settings in the initial flake evaluation it will drop your local cache, yes. If you don't, you will use your system cache settings. But I understand that this can be unwanted. Digging in the forum and on github, it seems that also `extra-substituters` is supported. I will push a version with those. That hopefully resolves this concern.

sheepforce: Version with `extra-substituters` pushed and working.

markuskowa:
Ok, let's give this version a try and see how it works out. Flakes are still experimental (even if treated like a stable feature by many people) and under development. Things may change in upcoming versions of Nix.

PR #185, opened 2022-04-28
bagel: fix build with gcc >= 10

markuskowa:
gcc and gfotran have been upgraded to version 10 and become stricter. Unbreak the build with a debichem patch.


sheepforce:
The patch is also directly available from upstream [here](https://github.com/qsimulate-open/bagel/commit/9ba401b237ca46a07c060b37016d79787ca39829), so that we wouldn't need to fetch debian patches.

markuskowa:
I overlooked that, which is probably due to the lack of useful commit messages. I will fix it in the next iteration. There is more fallout from the gcc/gfortran update.

Issue #186, opened 2022-04-29
Fix fallout from gcc/gfortran-10 update


markuskowa:
The latest update of gcc and gfortran to version 10 caused some packages to break due to stricter code checking:

* [x] bagel https://github.com/markuskowa/NixOS-QChem/pull/185
* [x] amd-scalapack
* [x] exatensor
* [x] dirac https://github.com/markuskowa/NixOS-QChem/pull/190
* [x] iboview  (probably needs a C++ patch to be gcc-10 compatible)
* [ ] molpro-ext (should be updated probably)
* [x] mcdth
* [x] mesa-qc https://github.com/markuskowa/NixOS-QChem/pull/188
* [x] molcas-18.09
* [x] molden (7.1 may solve the problems, direct fix: https://github.com/NixOS/nixpkgs/pull/172348)
* [x] orient
* [x] pegamoid
* [x] siesta  (upstream nixpkgs), see https://github.com/NixOS/nixpkgs/pull/172182
* [x] wfoverlap

See also [Porting to GCC 10](https://gcc.gnu.org/gcc-10/porting_to.html).

sheepforce:
I am currently a little bit short on time, but will try to fix mine as soon as I find a free time slot.

sheepforce:
I've tried also fixing Molden. The issue would probably simply be solved by switching to Molden7.0. Unfortunately fetchurl and curl in general refuse connecting to the site due to some SSL certificate errors. Also had no success with just switching to http there.

markuskowa:
Molden may require more work. I have tried to update it several times, but ran into a variety of issues (changes in the build system, segfaults in the final binary).
Molden 7.1 is now available (put online two days ago). I will give it another try.

markuskowa:
Lesson learned: some fixes broke the build against gfortran-9, but updating the nixpkgs pinning of course broke the rest that wasn't yet fixed. In the future, I will make a staging branch for such "major breakage" situations. PRs will then go on the staging branch first until everything is fixed. This will guarantee that we have a fully "green" master at all times.

PR #187, opened 2022-05-05
GCC Update Fixes

sheepforce:
This fixes some of the problems described in #186 

- `orient` fixed by upstream updates
- `wfoverlap` fixed by upstream updates
- `pegamoid` fixed by upstream updates, but other issues were introduced with setuptools, that I circumvent in the update
- `molcas1809` has probably no easy fix for this old version, except actually patching all type mismatches. I vote for just pinning gfortran to gfortran9. The hard reality is probably, that it is not viable to support outdated versions of software for new compiler versions forever. Although it would be sad to drop it entirely as this is the last supported version for SHARC, SHARC also runs fine with newer versions. Then again SHARC is also slowly dying as they refuse to update to Python3. I am open to just remove `molcas1809`
- `iboview` has a wrong type annotation *somewhere*. It basically constructs a set or dict somewhere from values, which are not declared as const. In the new libc++/glibc versions this is an error, as these ordered structures would be invalidated by intransparently changing values in them. Unfortunately I am completely lost in the mysteries C++ code and errors are to me and I have absolutely no idea where this would need to be fixed. It is probably just a `const` declaration missing at 2 or 3 positions.  The short term solution is to also pin the gcc version to 9 ... I am OK with it until upstream fixes it properly. What do you think?

### DIFF
--- a/overlay.nix
+++ b/overlay.nix
@@ -174,6 +174,7 @@ let
 
         molcas1809 = callPackage ./pkgs/apps/openmolcas/v18.09.nix {
           blas = final.blas-ilp64;
+          gfortran = final.gfortran9;
         };
 
         molcas = callPackage ./pkgs/apps/openmolcas/default.nix {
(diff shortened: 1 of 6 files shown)


sheepforce:
Hm, iboview was actually not fixed by switching back to GCC9 ... Now we have undefined references to glibcxx. :disappointed: My C++ is not good enough to fix this, I am afraid.

sheepforce:
Alrighty, have removed the iboview commits. 

I am sceptical with respect to IboView. The last release took 6 years, but let's hope for the best :)

PR #190, opened 2022-05-12
dirac: 21.0 -> 22.0

sheepforce:
This update includes cleanups of the derivation (not relying on the strange wrapper around CMake anymore), a license change to LGPL (yeaah :partying_face:), and fixes to compile with gfortran10 (this was quite some journey ...)


markuskowa:
Yay! One more fixed. Exatensor may potentially also be updated, but I did not dare to touch it.

PR #192, opened 2022-05-13
Use composeExtensions to preserve existing Python package overrides

### LINKED ISSUE #183: Workaround for Python `packageOverrides`
Thanks so much for this project! I've found it useful and have learned a lot from studying how it's organized.

I'm running into an issue that when I need to override packages in the `qchem.pkgs.python3` set, the overrides already applied by the overlay ([here](https://github.com/markuskowa/NixOS-QChem/blob/aff5d92a402dbb6fd72412dd93d2330c7a5e8b6b/overlay.nix#L287)) are apparently wiped out.

For example, I'd like to do something like the following:

```nix
{
  python3 = pkgs.qchem.python3.override {
    packageOverrides = prev: _: {
      mypackage = prev.buildPythonPackage {
        # …
      };
    };
  };

  pythonEnv = python3.withPkgs (ps: [ ps.mypackage ps.openmm ]);

  # …
}
```

But this fails with `error: attribute 'openmm' missing`.

I think this is related to the long-standing issue in nixpkgs: https://github.com/NixOS/nixpkgs/issues/44426

One of the [workarounds](https://github.com/NixOS/nixpkgs/issues/44426#issuecomment-410428253) mentioned in that issue is for overlays to expose `pythonOverrides` directly so that multiple overrides can be composed downstream using `lib.composeExtensions`. I wonder if you'd consider a PR exposing this in the overlay, or know of a different workaround?

mcwitt:
Resolves #183 

This PR addresses the scenario where the qchem overlay is applied after another overlay that overrides Python packages. Currently, the previously-applied overrides will be lost due to Python overrides not composing by default (https://github.com/NixOS/nixpkgs/issues/44426).

The following is a minimal reproduction:

```nix
let
  qchemOverlay = import ./overlay.nix;

  myOverlay = _: prev: {
    python3 = prev.python3.override {
      packageOverrides = (_: _: { foo = 42; });
    };
  };

  pkgs = import <nixpkgs> {
    overlays = [
      myOverlay
      qchemOverlay
    ];
  };
in
pkgs.qchem.python3.pkgs.foo
```
Currently, this expression fails to evaluate:
```
$ nix eval -f test.nix
error: attribute 'foo' missing
```

This PR modifies the qchem overlay to preserve existing overrides, using the variant of `override` that accepts a function from old to new overrides and `lib.composeExtensions`. This is similar to the approach used in [tweag/jupyterWith](https://github.com/tweag/jupyterWith/blob/93f75799b527b03de9acc09ae18cc4b8817fb0c5/nix/python-overlay.nix#L97-L102).

### DIFF
--- a/overlay.nix
+++ b/overlay.nix
@@ -281,8 +281,13 @@ let
         xtb = callPackage ./pkgs/apps/xtb { };
 
         ### Python packages
-        python3 = super.python3.override { packageOverrides = pythonOverrides cfg self super; };
-        python2 = super.python2.override { packageOverrides = pythonOverrides cfg self super; };
+        python3 = super.python3.override (old: {
+          packageOverrides = super.lib.composeExtensions (old.packageOverrides or (_: _: { })) (pythonOverrides cfg self super);
+        });
+
+        python2 = super.python2.override (old: {
+          packageOverrides = super.lib.composeExtensions (old.packageOverrides or (_: _: { })) (pythonOverrides cfg self super);
+        });
 
         #
         # Libraries


markuskowa:
> Currently, this expression fails to evaluate:
> 
> ```
> $ nix eval -f test.nix
> error: attribute 'foo' missing
> ```

nix-build test.nix works with this (e.g. use a real package):
```
let
  qchemOverlay = import ./overlay.nix;

  myOverlay = _: prev: {
    python3 = prev.python3.override {
      packageOverrides = (_: p: { foo = p.numpy; });
    };
  };

  pkgs = import <nixpkgs> {
    overlays = [
      myOverlay
      qchemOverlay
    ];
  };
in
pkgs.qchem.python3.pkgs.foo
```

mcwitt:
> nix-build test.nix works with this (e.g. use a real package):

@markuskowa ah, that's a more realistic reproduction!

Just to clarify, I assume you mean this works only on the PR branch? (Before the patch, this also fails for me, as expected)

markuskowa:
Yes, it works on the PR branch, and now also on master.

PR #199, opened 2022-06-10
iboview: 20211019 -> 20211019-RevA, gcc compatibility fixes

sheepforce:
@eljost found a patch that fixes the GCC compatibility issue. Meanwhile I've updated to the most recent version.


markuskowa:
Great! Thanks!

PR #201, opened 2022-06-17
Psi4 Update

sheepforce:
[Upstream Release](https://github.com/psi4/psi4/releases/tag/v1.6)

This is an update to the latest upstream release. it also adds support for the CPPE library with polarisable embedding and LibECPint for fast ECP intergrals.

Of course Psi4 has again changed how to build libint, so we are again rebuilding libint ...

### DIFF
--- a/overlay.nix
+++ b/overlay.nix
@@ -310,6 +310,8 @@ let
 
         amd-scalapack = callPackage ./pkgs/lib/amd-scalapack { };
 
+        libecpint = callPackage ./pkgs/lib/libecpint { };
+
         libefp = callPackage ./pkgs/lib/libefp { };
 
         libGDSII = callPackage ./pkgs/lib/libGDSII { };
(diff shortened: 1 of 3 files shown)


overlay.nix
```diff
@@ -310,6 +310,10 @@ let
 
         amd-scalapack = callPackage ./pkgs/lib/amd-scalapack { };
 
+        cppe = callPackage ./pkgs/lib/cppe { };
```
markuskowa, line 313: We already have CPP upstream (`<nixpkgs>/pkgs/development/libraries/science/chemistry/cppe/default.nix`)

markuskowa:
The `libint` is running since 2 days now and uses more than 17G RAM (`build_libint`). Not sure what caused this increase in build time.

sheepforce:
Same for me on my local machine ... This is still the code generator running. Psi4 moved a lot more of integrals into libint and removed their own code. They've added three-center integrals, F12 integrals and PCM integrals via libint and build higher order derivatives for all of them now. I'm curious how long their conda build takes. We could reduce to l=5 and make higher order l optional or so? But I guess this will later also hit us on nixpkgs ... :grimacing:

sheepforce:
Unfortunately not, no. The problem with excessive build times and incompatible configurations seems to be a problem for more people. There are some issues in libint's tracker. I will try another approach and use a prebuilt libint source: https://github.com/psi4/psi4/blob/master/external/upstream/libint2/CMakeLists.txt#L75

sheepforce:
OK, this at least visibly does _something_. Then let's keep l=5 with the pregenerated libint code for the moment and meanwhile I will have a look who else is able to do X2C calculations on Rhenium ... :(

sheepforce:
BTW @markuskowa have you successfully used the ADCC code with OpenBLAS? Even for relatively small molecules I am getting threading exceptions.

sheepforce:
Current version built and passed the tests :partying_face:

PR #204, opened 2022-07-19
janpa: init at 2.02

sheepforce:
Adds the Janpa program for natural population analysis and Molden basis function normalisation. This is our first Java program and I have never packaged Java before. I wasn't able to build from the sources (although it is open source), as at least as far as I understand Java, some kind of build instructions is missing. A Manifest.txt or something like this. Thus we have the binaries here.

Second issue is the license. It is open source, but I am not sure if we should list it as "free". It is basically a BSD3 license with the requirement to cite the academic papers on top. I guess we could call it free. It is not a more severe restriction than the GPL has with providing changes back, I would say.


markuskowa:
> Thus we have the binaries here.

Good enough for me. I have no idea either how to package JAVA.

PR #208, opened 2022-08-24
multiwfn: 3.7 -> 3.8

sheepforce:
An update to Multiwfn. This version finally is able to build with GFortran and droppped all the custom IFort extensions. A patch to build with gfortran for the Makefile is necessary.

Unfortunately this is declared the development version and no stable link seems to be provided. If a new dev version is uploaded it will break the hash. However, to cite the website:

> Latest formal version: 3.7　　Release date: 2020-Aug-14
(DO NOT use this currently, it is already quite old and no longer maintained!)

What do you think @markuskowa ? Should we update the hash whenever it breaks or wait for 3.8 to become stable?

### DIFF
--- a/pkgs/apps/multiwfn/default.nix
+++ b/pkgs/apps/multiwfn/default.nix
@@ -1,44 +1,36 @@
-{ stdenv, lib, makeWrapper, autoPatchelfHook, unzip,
-  writeScriptBin, fetchurl, glibc, xorg, libGL, motif
-} :
+{ stdenv, lib, makeWrapper, gfortran, unzip, fetchurl, xorg, libGL, motif, mkl }:
 
 stdenv.mkDerivation rec {
   pname = "multiwfn";
-  version = "3.7";
+  version = "3.8-2022-08-24";
 
-  # Only builds with Intel tooling. ifort is not available in Nix, therefore the binary package.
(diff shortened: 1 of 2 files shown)


pkgs/apps/multiwfn/default.nix
```diff
@@ -1,44 +1,36 @@
-{ stdenv, lib, makeWrapper, autoPatchelfHook, unzip,
-  writeScriptBin, fetchurl, glibc, xorg, libGL, motif
-} :
+{ stdenv, lib, makeWrapper, gfortran, unzip, fetchurl, xorg, libGL, motif, mkl }:
 
 stdenv.mkDerivation rec {
   pname = "multiwfn";
-  version = "3.7";
+  version = "3.8";
```
markuskowa, line 5:
```suggestion
  version = "3.8-2022-08-24";
```
It is inconvenient, when there is no stable release/link, but we can at least tag with a date. As long as you have a cached version, you will not even notice that the link broke.

sheepforce: OK, added the date :)

> As long as you have a cached version, you will not even notice that the link broke.

That is true and cachix should help there. But only for those who actually use it. So maybe someone else notices it first when it breaks :smile:

PR #214, opened 2022-10-05
multiwfn: update hash

sheepforce:


sheepforce:
> The date in `version` should probably also be updated.

Right, has been updated, too.

PR #217, opened 2022-10-06
mopac: init at 22.0.4

sheepforce:


pkgs/apps/mopac/default.nix
```diff
@@ -0,0 +1,35 @@
...
+    rev = "v${version}";
+    hash = "sha256-FhYJS61BIYB6wk2GQaIuxGTEXiXPC0JePCoY2JdRhMo=";
+  };
+
+  nativeBuildInputs = [
+    gfortran
+    cmake
+  ];
+
+  buildInputs = [ blas lapack ];
```
markuskowa, line 19: We probably need and assert check here for ILP64?

PR #219, opened 2022-10-14
dftd3: fix typo in license

sheepforce:


pkgs/apps/dft-d3/default.nix
```diff
@@ -24,6 +24,6 @@ stdenv.mkDerivation rec {
     description = "Dispersion correction for DFT";
     homepage = "https://github.com/dftbplus/dftd3-lib";
     platforms = platforms.unix;
-    license = licenses.gpl1Only;
+    license = licenses.gpl2Only;
```
markuskowa, line 27: It is not a typo, as far as I can see. The license file declares it as GPL version 1. https://github.com/dftbplus/dftd3-lib/blob/main/LICENSE

The readme file declares it as GPL v1 or later, in which case one define it as `gpl1Plus`.

sheepforce: Hm interesting. Yesterday I received an error on a Nix 2.11 installation with a pin from yesterday to Nixpkgs telling me that gpl1Only is an unkown license in Nixpkgs.

markuskowa: I fixed it yesterday from 'gpl1' to 'gpl1Only', since 'gpl1' (deprecated) was finally removed.
Maybe `gpl1Plus` is more appropriate.

sheepforce: Oh, I see. Then this is just a mistake of myself. I think we can close then.

PR #223, opened 2022-10-25
psi4: fix to libxc 5.2.3

markuskowa:
The update to [libxc-6.0.0](https://github.com/NixOS/nixpkgs/pull/197220) broke the psi4 build. In the 
Possible solution: fix the version in psi4 and remove this (ugly) hack when psi4 has an upgrade or fix available. 

CC @sheepforce: any other ideas?


sheepforce:
There is https://github.com/psi4/psi4/pull/1904 in Psi4 but already from 2020. :open_mouth: No better idea than fixing it for now and remove in Psi4 1.7.

PR #229, opened 2022-11-14
multiwfn: fix hash

sheepforce:


sheepforce:
Jup, this was a longer journey than expected. Two additional dependencies, a type error and a new makeflags. Works now.

PR #233, opened 2022-11-22
psi4: use upstream libint2

### LINKED ISSUE #226: Psi4 LibInt too low Angular Momentum
In #201 we have decided to stick to the pregenerated libint code from Psi4's modified libint version, basically the `5-4-3-6-5-4_mm4f12ob2` from [here](https://github.com/loriab/libint/releases).
Now after some time I've dealt with a 3d Nickel complex in Psi4 and the JK basis unfortunately already requires I functions. Being able to do at least 3d and 4d metals is important, I would argue. Thus we need _some_ solution to get the libint with AM=7 working again. @markuskowa any idea? Our last attempt hasn't even finished the code generation phase after two days with AM=7.

sheepforce:
closes #226 and reduces build time for Psi4 dramatically.


sheepforce:
A glorious. Compiles and finds libint2 from upstream but now wave functions are diverging. Will investigate. Possibly some shell ordering problem.

sheepforce:
The current version should build and run correctly with https://github.com/NixOS/nixpkgs/pull/202658

The build is fragile. Changing settings only slightly will make Psi4 segfault in certain modules, e.g. when calculating correlated gradients with density fitting methods. Our test case covers these. However, only after many many many hours of compilation :sweat_smile: 


Requires https://github.com/NixOS/nixpkgs/pull/202658 to be merged first. In `flake.nix` a working commit of nixpkgs is pinned currently.

sheepforce:
This version should build within ~ 2 hours and work as expected. Importantly even density fitting for gradients on Barium works, which is very nice. Should cover my metal problems now :)

sheepforce:
Now also includes an update to 1.7 (released just few days ago), can do gradients with density fitting and Douglas-Kroll on Barrium and MRCC and CFour are optionally available.

Currently tracks a special branch that includes both the libint and qcelemental PR s from nixpkgs. As soon as https://github.com/NixOS/nixpkgs/pull/202658 reaches nixpkgs-unstable we could merge?

markuskowa:
Yes, I think this looks good now. This PR is ready to go as soon as the upstream fix is available.

sheepforce:
libint has reached unstable. flake.lock is updated, can be merged.

markuskowa:
Ok, so far so good. Now we have the following problems:
* [`libint` timed out on Hydra](https://hydra.nixos.org/eval/1787373?filter=libint&compare=1787359&full=#tabs-still-fail). 
Did the build time of the new change significantly?
* The current nixpkgs-unstable removes `numpy` for python-2.7, which causes problems with SHARC. I will merge this  PR into staging instead.

markuskowa:
The `tests.psi4` fails now:
```
initPhase
setupPhase
Copying aux files
 input
runPhase
1..1
not ok 1 PSI4
# (in test file /nix/store/3jwfwdw44khjl70ji128vww87lyjj8qn-psi4.bats, line 22)
#   `grep "Final energy is    -76.235085" output' failed
# Optimizer: Optimization complete!
```

sheepforce:
> [libint timed out on Hydra](https://hydra.nixos.org/eval/1787373?filter=libint&compare=1787359&full=#tabs-still-fail).
Did the build time of the new change significantly?

Hm, this is strange. The code generation phase takes significantly more time than compilation on my machine and even then both together take slightly more than 2 hours. Could it be that hydra is building single-threaded?

> The current nixpkgs-unstable removes numpy for python-2.7, which causes problems with SHARC. I will merge this PR into staging instead.

Wow ... Needed to happen at some point. I would now simply consider Sharc broken if still not supporting Python3 after so many years. Just recently the Sharc developers told us that they are porting to Python3 (and rewriting parts), but it's not there yet.

> The tests.psi4 fails now

Completely forgot about the bats test. I've removed it. It is the same that is running in the `installCheckPhase` in Psi4 and just fails because OptKing (now external dep) changed printing. In the Psi4 expression itself it has been fixed.

sheepforce:
> Could it be that hydra is building single-threaded?

Not that bad but not great either. Just 2 cores are being used in the libint build:
```
build flags: -j2 SHELL=/nix/store/nfgpn9av331q7zi1dl6d5qpir60y513s-bash-5.1-p16/bin/bash
```

sheepforce:
Do we have any mechanism to influence it? It should build less ERIs than before the change. The only additional integrals are the multipole and one-electron integrals, which are fairly cheap compared to the ERIs. [Interestingly, in the last 2.7.1 build of libint it got much more cores: `build flags: -j16 SHELL=/nix/store/nfgpn9av331q7zi1dl6d5qpir60y513s-bash-5.1-p16/bin/bash`](https://hydra.nixos.org/build/200123508/nixlog/2)

PR #245, opened 2023-01-23
multiwfn: use external stable git as src and build instruction

### LINKED ISSUE #239: Multiwfn Hash and Repo
Currently the Multiwfn hash changes every week and the patches for CP2K and Multiwfn quickly outdate. I am thinking about forking it in a separate Git repository where a Pipeline automatically attempts an update every week. We could then point the derivation to this fork. It would also make patching easier. What is your opinion on this? The current situation is just super fragile.

sheepforce:
Closes #239 

I've created https://gitlab.com/theoretical-chemistry-jena/quantum-chemistry/multiwfn to solve the trouble with the breaking hashes and patches for the Multiwfn development version. The linked repo updates itself weekly and tries to port the gfortran patches for each update. The remote repo builds Multiwfn using Nix for its own pipeline and I am simply reusing the build instructions here.


overlay.nix
```diff
@@ -221,7 +221,14 @@ let
...
 
-        multiwfn = callPackage ./pkgs/apps/multiwfn { };
+        multiwfn = let
+          src = prev.fetchFromGitLab {
+            owner = "theoretical-chemistry-jena/quantum-chemistry";
+            repo = "multiwfn";
+            rev = "16cdd7a9a5bf4701e69b16c818bf8ccb2476ce5a";
+            hash = "sha256-frmAHQdoic84Ekwd7IrBUVql6PMzyN8YD/PZuiDGCuo=";
+          };
+        in callPackage "${src}/nix/multiwfn.nix" { };
```
markuskowa, line 231: I would prefer to keep the Nix expression in this repository, instead of pulling it from your Gitlab first. I think that makes it more transparent of how the software is packaged.
One other disadvantage: in this form a `fetchFromGitLab` is performed in the eval stage instead of the package's build phase. I.e. just calling `nix-instantiate release.nix` triggers a  fetchFromGitLab`.

sheepforce: I guess this is our local replication of the flakes RFC discussion and whether to split up Nixpkgs into multiple repos :laughing: 

We already had this issue with pysisyphus, where the expression is much more complex. I was not such a big fan of the replicated code. But there are certainly good reasons for both sides.

The current state is a foul compromise, I would say. My preferred solution would actually be to use Multiwfn and and pysisyphus as flake inputs and expose them in the overlay again. What do you think about this option? I know that it still requires the fetch in the instantiate phase but is this a real problem? The repos are small and cache nicely.

markuskowa: I would go with the traditional way of having the derivation in the overlay. Pulling it from remote is "emulating flakes" in a not-clean way. In a non-flake setting, you still have a synchronization issue: you pull a derivation from remote that relies on certain inputs that are provided by the overlay. That creates an invisible dependency that is much harder to track and fix (e.g. name of an input is changed) than when the derivation is directly in the overlay. If you have no control over the "input source", that will leave you in a deadlock situation, which needs to be patched in another way again.
Keep in mind, that your input and the overlay need to use the same nixpkgs version for this to work reliably. 

In principle, I am against relying on flakes at the moment. Flakes right now are an experimental feature and have not been stabilized yet (even though many people use them). I do not use flakes at all on my side, but rely on the traditional Nix cli interface (not even new "nix x" interface is declared stable by the way). One big showstopper is the lack of a nix-shell/shebang replacement, which crucial for a convenient integration with Slurm. I am also providing these packages via Nix to a non-expert user base. Doing this with an interface that can change its behavior in subtle ways twice a year sounds like a major headache.

There will come a time when flakes are stable and everything gets moved over to flakes, but for now I personally see flakes as an optional feature.

Here are my 5 cents on the "breaking nixpkgs up" discussion: I can see why this may look appealing over the mono-repo situation. A smaller nixpkgs-core would mean things can be changed more easily. The breakage will then not happen in nixpkgs, but still somewhere else. I still have a hard time imagining how this scenario would really look like. The things I am most worried about are:
* Instead of getting all my dependencies from one well-curated source (nixpkgs), I will have to go shopping from different repos that may or may not be well maintained. I have to judge the quality of each of these, potentially many, sources separately. 
* How many flake inputs will I depend on directly and indirectly? Let's say for example 50 (direct or indirect). Each one of these will probably depend on `nixpkgs-core`. Worst case scenario, I will need to pull 50 different versions of `nixpkgs-core`? That means that also the number of build products in the store may multiply and even small packages will explode in size?
* There are a lot of packages in nixpkgs that do not have a maintainer, but are still community maintained when needed. Most of these packages may then just disappear?
* What security/reliability implications will this situation have? The JS/npm(?) community has such a distributed scenario, where many projects have relied on a crucial, but small package that was suddenly removed (or had other problems) and caused major issues.

sheepforce: Those are all valid points and I have adapted it to have the nix expression locally in this repo.

A few thoughts on particular points, anyway. While the stability of flakes is truly questionable yet, especially the non-expert users here seem to deal with flakes much better thanks to the more consistent interface and more static rules how things should look like. A default.nix with imports of pins from other files or niv is not as easy as a flake update. In our own projects (Spicy, Pysisyphus mainly), we went for flakes and use the flake-compat tools to have a shell.nix and default.nix if this backwards compatibility is required. For the last two years this never broke. While they might not be stable, they are stable enough, I would say.
The missing nix-shell replacement is truly a shame, I have to confess.

PR #251, opened 2023-02-13
overlay: unbreak optimized stdenv

markuskowa:
https://github.com/NixOS/nixpkgs/pull/215056 broke our, not so clean, method to add compiler flags to `stdenv`.
The new version uses proper functions from `nixpkgs` to modify `stdenv`.


markuskowa:
Now I remember why I did not use that solution in the first place: it appears to some packages as if we are cross compiling for a new architecture, creating a whole new array of problems.

PR #257, opened 2023-02-24
Infrastructure cleanup

sheepforce:
This PR is meant to address some challenges of infrastructural nature and meant to clean up a few things and update a few of our workflows.

# Goals
* [x] Make the flake pass all checks of `nix flake check`
* [x] Provide development tools to write work with the overlay in the `devShells` output attribute, including direnv integration
* [x] New and stable mechanism to find changed packages
* [x] Build changed packages via GitHub actions
* [ ] ~~Push build artifacts from the GitHub actions to Cachix~~
* [ ] Integrate on-site GitHub runner with more powerful hardware and local cache

# Open Questions
* On-site GitHub runners could build and test proprietary packages we would have access to, provided we use the `srcurl` mechanism and the GitHub runner machine has access to that respective host. In our setup this should be possible without problems. Our GitHub runner machine would be in the same subnet as the file server. However, I am not entirely sure if we _should_ do this. It would require some protection mechanism of who is allowed to trigger pipelines.
* Currently, we are using the `null` overlay to disable some proprietary packages in the flake. Is this strictly necessary? I think it does no harm exposing common unfree packages. To avoid their tests failing I would rather filter the tests?

### DIFF
--- a/.github/workflows/changedPaths.nix
+++ b/.github/workflows/changedPaths.nix
@@ -20,12 +20,12 @@ let
 
   # Recursive, flattened difference of attribute sets.
   recAttrSetDiff = l: r: lib.attrsets.filterAttrsRecursive
-       (lK: lV: !(builtins.elem lK auxIgnore) && (if lib.hasAttr lK r
-         then lV != r."${lK}"
-         else true)
-       )
-       l
-      ;
+    (lK: lV: !(builtins.elem lK auxIgnore) && (if lib.hasAttr lK r
+    then lV != r."${lK}"
+    else true)
+    )
+    l
+  ;
 
   # Simplify package set
   simplify = as: with lib.attrsets; mapAttrs (_: v: v.drvPath) (filterAttrs (_: v: isDerivation v) as);
@@ -48,7 +48,9 @@ let
   changedPyDerivs = recAttrSetDiff updPyDerivs basPyDerivs;
 
   # For CI purposes we only require the changed attribute keys.
-in { topLevel = builtins.attrNames changedDerivs;
-     python3 = builtins.attrNames changedPyDerivs;
-   }
+in
+{
+  topLevel = builtins.attrNames changedDerivs;
+  python3 = builtins.attrNames changedPyDerivs;
+}
 
--- a/.github/workflows/changedPaths.nix
+++ b/.github/workflows/changedPaths.nix
@@ -12,7 +12,7 @@ let
 
   pkgs = import (builtins.fetchTarball "https://github.com/NixOS/nixpkgs/archive/refs/heads/nixos-22.11.tar.gz") { };
 
-  lib = pkgs.lib;
+  inherit (pkgs) lib;
 
   # Ignored outputs, to be removed additionally from the attribute set.
   auxIgnore = [
--- a/.github/workflows/changedPaths.nix
+++ b/.github/workflows/changedPaths.nix
@@ -1,0 +1,54 @@
+updPath: basePath:
+
+let
+  releaseOpts = {
+    config = { };
+    allowUnfree = true;
+    preOverlays = [ ];
+    postOverlays = [ ];
+    buildVariants = false;
+    pin = true;
+  };
+
+  lib = (import "${basePath}/nixpkgs-pin.nix" { }).lib;
+
+  # Ignored outputs, to be removed additionally from the attribute set.
+  auxIgnore = [
+    "channel"
+    "nixexprs"
+  ];
+
+  # Recursive, flattened difference of attribute sets.
+  recAttrSetDiff = l: r: lib.attrsets.filterAttrsRecursive
+       (lK: lV: !(builtins.elem lK auxIgnore) && (if lib.hasAttr lK r
+         then lV != r."${lK}"
+         else true)
+       )
+       l
+      ;
+
+  # Simplify package set
+  simplify = as: with lib.attrsets; mapAttrs (_: v: v.drvPath) (filterAttrs (_: v: isDerivation v) as);
+
+  # Get packages from the updated and base versions
+  updPkgs = (import "${updPath}/release.nix" releaseOpts).qchem;
+  basPkgs = (import "${basePath}/release.nix" releaseOpts).qchem;
+
+  # We are not interested in the realised packages, the derivation paths are
+  # enough and save a lot of time.
+  updDerivs = simplify updPkgs;
+  basDerivs = simplify basPkgs;
+
+  # Handle python packages separately to avoid recursion errors with filterAttrsRecursive
+  updPyDerivs = simplify updPkgs.python3;
+  basPyDerivs = simplify basPkgs.python3;
+
+  # Changed derivations in update
+  changedDerivs = recAttrSetDiff updDerivs basDerivs;
+  changedPyDerivs = recAttrSetDiff updPyDerivs basPyDerivs;
+
+  # For CI purposes we only require the changed attribute keys.
+in { topLevel = builtins.attrNames changedDerivs;
+     python3 = builtins.attrNames changedPyDerivs;
+   }
+
--- a/.github/workflows/changedPaths.nix
+++ b/.github/workflows/changedPaths.nix
@@ -10,7 +10,7 @@ let
     pin = true;
   };
 
-  lib = (import "${basePath}/nixpkgs-pin.nix" { }).lib;
+  lib = (import "${basePath}/nixpkgs-pin.nix" (import <nixpkgs> {})).lib;
 
   # Ignored outputs, to be removed additionally from the attribute set.
   auxIgnore = [
--- a/.github/workflows/changedPaths.nix
+++ b/.github/workflows/changedPaths.nix
@@ -10,7 +10,9 @@ let
     pin = true;
   };
 
-  lib = (import "${basePath}/nixpkgs-pin.nix" (import <nixpkgs> {})).lib;
+  pkgs = import (builtins.fetchTarball "https://github.com/NixOS/nixpkgs/archive/refs/heads/nixos-22.11.tar.gz") { };
+
+  lib = pkgs.lib;
 
   # Ignored outputs, to be removed additionally from the attribute set.
   auxIgnore = [
--- a/.github/workflows/editorconfig.yml
+++ b/.github/workflows/editorconfig.yml
@@ -12,6 +12,6 @@ jobs:
       uses: tj-actions/changed-files@v35
     - uses: cachix/install-nix-action@v20
     - run: |
-      for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+        for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
           nix run nixpkgs#editorconfig-checker -- $file
         done
--- a/.github/workflows/editorconfig.yml
+++ b/.github/workflows/editorconfig.yml
@@ -15,8 +15,4 @@ jobs:
     - uses: workflow/nix-shell-action@v3
       with:
         packages: editorconfig-checker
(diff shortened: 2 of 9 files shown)


markuskowa:
A few comments here:
*  Make the flake pass all checks of `nix flake check`: that may require to move the package set to `legacyPackages` as `packages` does not allow for subsets.
* `devShells` output attribute, including direnv integration: `direnv` is a nice-to-have add on, but I would not make it mandatory.
* Build changed packages via GitHub actions: the mechanism to find changed packages via the GitHub action is currently broken. This is, btw. not as trivial as I thought. OfBorg puts more effort in to figure out what exactly has changed.
* Push build artifacts from the GitHub actions to Cachix: I would try to save space here instead and either open another Cachix account for exclusively the GitHub runner or rather provide a local cache.

sheepforce:
> * Make the flake pass all checks of `nix flake check`: that may require to move the package set to `legacyPackages` as `packages` does not allow for subsets.

Yeah, the python package set is not exported in packages, that's true. However, I would argue that this is fine. Python packages used as executables are exported via `toPythonApplication` in the `qchem` set so that they are accessible. If a python package is meant to be used as a library instead, the better way is to access it via the overlay function so that the python versions don't become a mess. The current state already passes `nix flake check`.

> devShells output attribute, including direnv integration: direnv is a nice-to-have add on, but I would not make it mandatory.

It is not. If direnv is not installed it won't matter. If it is installed and not to be used it is enough to just not run `direnv allow`. 

> Build changed packages via GitHub actions: the mechanism to find changed packages via the GitHub action is currently broken. This is, btw. not as trivial as I thought. OfBorg puts more effort in to figure out what exactly has changed.

Yep, I've also noticed. I think I can imagine another solution. I will try it out soonish, I hope. I've also hat a look at nixpkgs update to save us from all manual package upgrades. Unfortunately it quite tied to the nixpkgs repo and also quite fancy Haskell with effect systems and that stuff ... :grimacing: 

> Push build artifacts from the GitHub actions to Cachix: I would try to save space here instead and either open another Cachix account for exclusively the GitHub runner or rather provide a local cache.

I guess this point has to aspects: pushing into the current cachix for "end-users" which we could restrict to commits on master and the pinned nixpkgs. And the second one is accelerating the builds for local runners. Here I agree, I would also prefer a local cache.

markuskowa:
> I guess this point has to aspects: pushing into the current cachix for "end-users" which we could restrict to commits on master and the pinned nixpkgs. And the second one is accelerating the builds for local runners. Here I agree, I would also prefer a local cache.

Hydra does that already for the pinned master branch and the release branches. We can leave it like that. I think the GitHub runners are a good option to handle the PRs.

sheepforce:
I am really struggling with the GitHub actions. They are lacking the feature to create dependent child pipelines. In our case I would have created a stage 1 pipeline, that figures our which packages have changed. The stage 1 pipeline would then create a number of stage 2 pipelines, that build each package. Unfortunately it is extremely cumbersome to achieve something like this with GitHub actions. However, I am regularly using this in GitLab CI and would just register connect a  GitLab CI-repo with this repo. Then the GitLab-CI would provide the results via a Webhook, similar to what ofborg is doing.

markuskowa:
> I am really struggling with the GitHub actions. They are lacking the feature to create dependent child pipelines. In our case I would have created a stage 1 pipeline, that figures our which packages have changed. The stage 1 pipeline would then create a number of stage 2 pipelines, that build each package. Unfortunately it is extremely cumbersome to achieve something like this with GitHub actions. However, I am regularly using this in GitLab CI and would just connect a GitLab CI-repo with this repo. Then the GitLab-CI would provide the results via a Webhook, similar to what ofborg is doing.

I am not familiar with the GitLab-CI. I think in GitHub actions, each action is completely isolated. The simplest solution is probably to do both steps in one action? I  am not sure how to share information between actions (but there should be a way to achieve this)
Did you manage to find out the packages that changed?

sheepforce:
> I am not sure how to share information between actions (but there should be a way to achieve this)

I think there is, but this is not even what I am missing. I want a dynamic number of pipelines in the second stage. But yes, we can also put everything into one giant monolithic pipeline. 

>  Did you manage to find out the packages that changed?

No, not yet, unfortunately. The idea is quite simple and robust, but I am struggling with implementing it and always end in an infinite recursion, that I do not understand. The idea is to create a nix expression in a file `getDiff.nix` that roughly does the following:

```nix
basePath: newPath:

let
  flakeTools = ... # import the flake-compat stuff, that provides lib functions to create package sets from flake.nix and flake.lock
  basePkgs = (flakeTools { src = basePath; }).defaultNix.packages.x86_64-linux;
 
```

sheepforce:
Haha, got it! :blush: :partying_face: 

```nix
updPath: basePath:

let
  releaseOpts = {
    config = { };
    allowUnfree = true;
    preOverlays = [ ];
    postOverlays = [ ];
    buildVariants = false;
    pin = true;
  };

  lib = (import "${basePath}/nixpkgs-pin.nix" { }).lib;

  # Get packages from the updated and base versions
  updPkgs = (import "${updPath}/release.nix" releaseOpts).qchem;
  basPkgs = (import "${basePath}/release.nix" releaseOpts).qchem;

  # We are not interested in the realised packages, the derivation paths are
  # enough and save a lot of time.
  updDerivs = with lib.attrsets; mapAttrs (_: v: v.drvPath) (filterAttrs (_: v: isDerivation v) updPkgs);
  basDerivs = with lib.attrsets; mapAttrs (_: v: v.drvPath) (filterAttrs (_: v: isDerivation v) basPkgs);

  # Ignored outputs, to be removed additionally from the attribute set.
  auxIgnore = [
    "channel"
    "nixexpr"
  ];

  # Recursive, flattened difference of attribute sets.
  recAttrSetDiff = l: r: lib.attrsets.filterAttrs
       (lK: lV: !(builtins.elem lK auxIgnore) && (if lib.hasAttr lK r
         then lV != r."${lK}"
         else true)
       )
       l
      ;

  # Changed derivations in update
  changedDerivs = recAttrSetDiff updDerivs basDerivs;

  # For CI purposes we only require the changed attribute keys.
in builtins.attrNames changedDerivs
```

And to be called as `nix eval --impure --expr 'import ./.github/workflows/changedPaths.nix $PATH_TO_PR $PATH_TO_MASTER`.

sheepforce:
OK, done with what I wanted to do, except the registration of our custom runner. However, for the actual code changes that does not matter for now.

The rebuilds seem to work nicely as far as I can tell, for an example see here: https://github.com/sheepforce/NixOS-QChem/actions/runs/4354792813

However, the mechanism will only work once this PR is merged, as the workflow requires some of my changes here.

sheepforce:
Hm, a problem for the checkout is a cross-fork PR, I will try to figure out how to fix.

sheepforce:
So the `rebuilds / Pinned` legitimately fails as the CI requires my changes to be in the PR target branch (`master`).
`rebuilds / Master` and `rebuilds / Unstable` also fail for the same reason, but would also still fail due to #262 .

`nix flake check` is interesting, however. This one works flawlessly on my NixOS machine and I suspect the version difference (2.11 vs 2.14) to be responsible. Do you have more insight, @markuskowa ?

sheepforce:
I have rebased onto staging (and target staging as base). However, the stdenv fix is still making trouble;

```bash
$ nix flake check
warning: Using saved setting for 'allow-import-from-derivation = true' from ~/.local/share/nix/trusted-settings.json.
warning: Using saved setting for 'extra-substituters = https://nix-qchem.cachix.org' from ~/.local/share/nix/trusted-settings.json.
error: The ‘env’ attribute set cannot contain any attributes passed to derivation. The following attributes are overlapping: NIX_CFLAGS_COMPILE
(use '--show-trace' to show detailed location information)
```

nixpkgs-pin.nix
```diff
@@ -1,10 +1,14 @@
-pkgs:
-
 let
-  # tested nixpgks-unstable version
-  nixpkgsGH = with builtins; (fromJSON (readFile ./flake.lock)).nodes.nixpkgs.locked;
+  defaultNix = (import
+    (
+      let lock = builtins.fromJSON (builtins.readFile ./flake.lock); in
+      builtins.fetchGit {
```
markuskowa, line 5: Found the following trick here:
https://github.com/ZilchOS/bootstrap-from-tcc/blob/de68f791590dbc48c7f3e738edd086fbe95e7b4b/default.nix#L52

Define a minimal `fetchurl`
```
  fetchurl = { url, sha256 }: derivation {
    name = builtins.baseNameOf url;
    inherit url;
    urls = [ url ];
    unpack = false;

    builder = "builtin:fetchurl";
    system = "builtin";
    outputHashMode = "flat"; outputHashAlgo = "sha256";
    preferLocalBuild = true;
    outputHash = sha256;
  };
```

sheepforce: Hm, unfortunately this does not work as it results in a hash mismatch. I guess this is due to `unpack = false`. However, `unpack = true` seems to only work with NARs.

sheepforce:
@markuskowa Here we are. Everything should work now, the checks even work now without being in the base branch of the PR, nixpkgs-pin.nix is untouched and thus still working. Please have a look :)

.github/workflows/editorconfig.yml
```diff
@@ -6,17 +6,12 @@ jobs:
...
-    - if: ${{ failure() }}
-      run: |
-        echo "::error :: Hey! It looks like your changes don't follow our editorconfig settings."
+    - uses: actions/checkout@v3
+    - name: get changed files
+      id: changed-files
+      uses: tj-actions/changed-files@v35
+    - uses: cachix/install-nix-action@v20
+    - run: |
+      for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
```
markuskowa, line 15: There seems to be a YAML syntax error here (see "actions"). Maybe this needs to be indented?

.github/workflows/changedPaths.nix
```diff
@@ -0,0 +1,58 @@
...
+    allowUnfree = true;
+    preOverlays = [ ];
+    postOverlays = [ ];
+    buildVariants = false;
+    pin = true;
+  };
+
+  pkgs = import (builtins.fetchTarball "https://github.com/NixOS/nixpkgs/archive/refs/heads/nixos-22.11.tar.gz") { };
+
+  lib = pkgs.lib;
```
markuskowa, line 15: Statix suggestion:
```suggestion
  inherit (pkgs) lib;
```

.github/workflows/changedPaths.nix
```diff
@@ -0,0 +1,58 @@
...
+  recAttrSetDiff = l: r: lib.attrsets.filterAttrsRecursive
+    (lK: lV: !(builtins.elem lK auxIgnore) && (if lib.hasAttr lK r
+    then lV != r."${lK}"
+    else true)
+    )
+    l
+  ;
+
+  # Simplify package set
+  simplify = as: with lib.attrsets; mapAttrs (_: v: v.drvPath) (filterAttrs (_: v: isDerivation v) as);
```
markuskowa, line 33: Statix suggestion ("Found eta-reduction"):
```suggestion
  simplify = as: with lib.attrsets; mapAttrs (_: v: v.drvPath) (filterAttrs (_: isDerivation) as);
```

markuskowa (commented):
Very nice! This is certainly an improvement.

.github/workflows/changedPaths.nix
sheepforce: Ah my old enemy, the eta reduction ^^ Hlint for haskell does the same to me and I tend to disagree. I don't like point free functions. I would prefer to keep this as is.

PR #266, opened 2023-03-17
Merge staging

markuskowa:
Ready for a clean master branch.

### DIFF
--- a/.github/workflows/nix-rebuild.yml
+++ b/.github/workflows/nix-rebuild.yml
@@ -1,47 +1,126 @@
-name: nix rebuilds
+name: rebuilds
 on: [pull_request]
 
 jobs:
-  run:
-    name: Calculate changed outputs
+  # Calculate changed derivations
+  get-changed-derivs:
+    name: Get changed derivations
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-      with:
-        path: pr
-    - uses: actions/checkout@v2
-      with:
-        path: master
-        ref: master
-    - uses: cachix/install-nix-action@v16
-      with:
-        nix_path: nixpkgs=channel:nixos-unstable
-    - name: Changed paths
-      run: |
-        cat /etc/nix/nix.conf
-        id
-        echo $NIX_PATH
-        nix-channel --add https://nixos.org/channels/nixos-21.11 nixpkgs
-        nix-channel --update
-        nix-instantiate ./pr/release.nix -A qchem
-        nix-env -f ./pr/release.nix -qaP --no-name --out-path --show-trace \
-          --option binary-caches 'https://cache.nixos.org/ https://nix-qchem.cachix.org/' \
-          --option trusted-public-keys '
-            nix-qchem.cachix.org-1:ZjRh1PosWRj7qf3eukj4IxjhyXx6ZwJbXvvFk3o3Eos=
-            cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
-          ' --arg allowUnfree false -A qchem | sort > paths-pr
-        nix-instantiate ./master/release.nix -A qchem
-        nix-env -f ./master/release.nix -qaP --no-name --out-path --show-trace \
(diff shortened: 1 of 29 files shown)


markuskowa:
Thanks for fixing `gau2grid`. This `buildPythonPackage`/`mkDerivation` seesm to be a rats nest. It seems like that the GitHub work flow treats warnings as errors?

.github/workflows/nix-rebuild.yml
```diff
@@ -1,47 +1,126 @@
...
+          fi
+      # Rebuild all changed top-level packages
+      - run: |
+          if [ -n "$(jq '.topLevel[]' chDerivs.json)" ]; then
+            nix build --accept-flake-config --override-input nixpkgs nixpkgs/nixpkgs-unstable --print-build-logs $(for p in $(jq '.topLevel[]' chDerivs.json | tr -d "\""); do echo ".#${p}"; done)
+          fi
+
+  # Rebuild changed derivations for nixpkgs/master
+  rebuild-master:
+    name: Master
```
markuskowa, line 101: I actually wonder if we really should build PRs against nixpkgs-master. The master branch itself contains a lot of temporary breakage and large rebuild sets, that may not have been built by Hydra. I do not even build locally against nixpkgs-master as this would result in a lot of irrelevant build failures. I think that building against nixpkgs-unstable completely covers our use-case of projecting ahead.
Would it maybe make more sense to build the merged PR against the pinned nixpkgs here?

sheepforce: I agree, the master rebuild could be removed.

>  Would it maybe make more sense to build the merged PR against the pinned nixpkgs here?

What exactly do you mean? :) Not building the head of the PR but the result after the merge commit?

markuskowa:
> Not building the head of the PR but the result after the merge commit?

Yes, build the result of the merge. (I.e. what would happen if one were to merge a PR).

sheepforce:
> Thanks for fixing `gau2grid`. This `buildPythonPackage`/`mkDerivation` seems to be a rat's nest.

Yeah, indeed. But luckily the fix is simple. I am surprised that this didn't bite us earlier.

> It seems like that the GitHub work flow treats warnings as errors?

GitHub actions simply use the exit code. This is Nix deciding to throw an exit code /= 0 and GitHub just checking this one. I think that is a reasonable strategy.
it might very well be that Nix 2.14 got stricter with this and many other things. For example, import from derivation is now always an error in Nix 2.14 with flake check. In consequence, everything that calls cabal2Nix is now an flake error :smiling_face_with_tear:

sheepforce:
Another issue we have in the pipelines is that they are attempting to build unfree packages, which of course fails. I will try to filter out unfree packages from the changed derivations.

.github/workflows/nix-rebuild.yml
sheepforce: > Yes, build the result of the merge. (I.e. what would happen if one were to merge a PR).

Done in https://github.com/Nix-QChem/NixOS-QChem/pull/266/commits/69038e31debd40a39d08f1e74d1557381bbf6223

In GitHub action I find this to be quite obscure ...

markuskowa:
> GitHub actions simply use the exit code. This is Nix deciding to throw an exit code /= 0 and GitHub just checking this one. I think that is a reasonable strategy.
> it might very well be that Nix 2.14 got stricter with this and many other things. For example, import from derivation is now always an error in Nix 2.14 with flake check. In consequence, everything that calls cabal2Nix is now an flake error

That is a bit concerning that errors may now depend on the Nix version. That takes away some reproducibility. My Hydra runs 2.11.1, which builds everything just fine.

sheepforce:
> That is a bit concerning that errors may now depend on the Nix version. That takes away some reproducibility. My Hydra runs 2.11.1, which builds everything just fine.

Agreed, these changes unfortunately invalidate all my Haskell expression. Their canonical way is to use `callCabal2nix` which intrinsically is an import from derivation.
But I guess somehow we should deal with it as correctly as possible.
We can't stop the Nix ecosystem to move forward :grimacing:

markuskowa:
ok, then we merge it now and fix the rest later. All my Hydras are green on the PR, i.e. nothing is really broken.

PR #267, opened 2023-03-20
Tblite cleanup

sheepforce:
This removes the entire tblite ecosystem from the overlay, as it is now part of nixpkgs.


markuskowa:
Can we maybe wait with this PR. I would like to merge the current staging branch first, before we update `flake.lock` again.
I.e. put this PR in the next staging round.

sheepforce:
Sure, let's wait then :)

markuskowa:
That was not intended.  It think you have to rebase it onto a branch that still exists, i.e. staging.

PR #269, opened 2023-03-22
update package list

markuskowa:
A task that screams "automation!"

### DIFF
--- a/.github/workflows/typos.yml
+++ b/.github/workflows/typos.yml
@@ -13,6 +13,6 @@ jobs:
     - uses: cachix/install-nix-action@v20
     - run: |
         for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
-          nix run nixpkgs#typos -- $(echo "${{ steps.files.outputs.all }}" | tr ' ' '\n' | sed 's:\(.*\):./\1:;/\.patch$/d' | tr '\n' ' ')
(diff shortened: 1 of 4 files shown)


sheepforce:
For JSON this should be easy using `builtins.toJSON`, shouldn't it? For Markdown I could think of a string generator in Nix.

markuskowa:
> For JSON this should be easy using `builtins.toJSON`, shouldn't it? For Markdown I could think of a string generator in Nix.

`scripts/package_list` already generates the list, but I still run it manually. A Github Action, that runs it commits the new list (if a different list was generated), would be nice.

sheepforce:
Hm, I dislike pipelines pushing and commiting. I would instead prefer to generate them as artifacts in the pipeline and link them from the README. This has the disadvantage to diverge for older commits at some point but hm, nicer than automatic commits I would say.

markuskowa:
> Hm, I dislike pipelines pushing and commiting. 

It could either add a commit with the updated list to a PR or be triggered once a PR is merged. The latter would be less intrusive for the PR workflow.

sheepforce:
Hm, yes but it would require the pipeline to be allowed to push to `master`. Any push will require a bot account or API key and secret management.

markuskowa:
So, the spell check should now create less noise.
I will see if I can come with an idea for GH action that can manage such an update task without being annoying (it needs to be easier and simpler than it is now).

PR #270, opened 2023-03-22
github/nix-rebuild: add names to job steps

markuskowa:
Let's try if we can add names to the job steps. That would make the runner output more readable.

Changes:
* Added naming to the job steps to increase readability in on PR page and actions page (detailed listing).
* The default of the checkout action is to check out the merge commit. HEAD is now HEAD of the PR and Merge is the merge commit.
*  Removed the nixpkgs-unstable rebuild job. This is a projection into the future that may be nice to have, but not sure if it's worth the resource usage.

### DIFF
--- a/.github/workflows/nix-rebuild.yml
+++ b/.github/workflows/nix-rebuild.yml
@@ -69,35 +69,6 @@ jobs:
             nix build --accept-flake-config --print-build-logs $(for p in $(jq '.topLevel[]' chDerivs.json | tr -d "\""); do echo ".#${p}"; done)
           fi
 
-  # Rebuild changed derivations for nixpkgs-unstable
-  rebuild-unstable:
-    name: Unstable
-    needs: get-changed-derivs
-    runs-on: ubuntu-latest
-    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/nix-rebuild.yml
```diff
@@ -110,20 +83,17 @@ jobs:
           name: nix-qchem
       # Checkout merge commit as if PR would already be merged
       - uses: actions/checkout@v3
-        with:
-          ref: ${{ github.sha }}
-          repository: ${{ github.event.pull_request.base.repo.full_name }}
```
sheepforce, line 115: There is a good chance that this won't work for cross-fork PRs. The `repository` line was explicitly required to fix this.

markuskowa: I opened a cross fork PR here: https://github.com/Nix-QChem/NixOS-QChem/pull/272
It looks like the checkout action is doing what it is supposed to do:
Check under "Checking out the ref". It says : "/usr/bin/git checkout --progress --force refs/remotes/pull/272/merge".

PR #271, opened 2023-03-23
python.vermouth/polyply: fix build

markuskowa:
Remove constraint for networkx 2.0. `networkx` has been updated to version 3.0 (https://github.com/NixOS/nixpkgs/pull/218614).
Both packages just build fine and polyply passes all tests.
@sheepforce can you check if that fix is okay?


sheepforce:
Looks good to me. If the interface is still compatible (which it apparently is in the relevant parts), I see no reason to expect a breakage.

PR #273, opened 2023-03-28
Merge staging

markuskowa:
This is an attempt to update flake.lock. GPAW and Vermouth are fixed now, but Psi4 now segfaults in the check phase.

TODO:
* Find the change in upstream nixpkgs that broke Psi4.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.github/workflows/changedPaths.nix
+++ b/.github/workflows/changedPaths.nix
@@ -1,0 +1,73 @@
+updPath: basePath:
+
+let
+  releaseOpts = {
+    config = { };
+    allowUnfree = false;
+    preOverlays = [ ];
+    postOverlays = [
+      (final: prev: {
+        turbomole = null;
+        cefine = null;
+        cfour = null;
+        gamess-us = null;
+        mrcc = null;
+        orca = null;
+        qdng = null;
+        vmd = null;
+        mesa-qc = null;
+        mcdth = null;
+        nixGL = null;
+      })
+    ];
+    buildVariants = false;
+    pin = true;
+  };
+
+  pkgs = import (builtins.fetchTarball "https://github.com/NixOS/nixpkgs/archive/refs/heads/nixos-22.11.tar.gz") { };
+
+  inherit (pkgs) lib;
+
+  # Ignored outputs, to be removed additionally from the attribute set.
+  auxIgnore = [
+    "channel"
+    "nixexprs"
+    "tested"
+  ];
+
+  # Recursive, flattened difference of attribute sets.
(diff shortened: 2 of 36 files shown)


sheepforce:
I found it but it is awkward. It is the BaF2 example failing in the calculation of the dispersion correction. If one calculates the energy and not a gradient via finite differences, there is actually a helpful message, complaining about dftd3 being too old. This is a qcengine upstream update, that now checks for this version boundary.

I will see what I can do.

sheepforce:
OK, it is quite bad. QCEngine checks for at least dftd3 3.2.1. At https://www.chemie.uni-bonn.de/grimme/de/software/dft-d3/get_dft-d3 there is not even a version 3.2.1, the download link is unstable and older versions are not accessible.
QCEngine and Psi4 get a binary version from their own Conda Forge channel here https://anaconda.org/psi4/dftd3/files

I have no idea where Psi4 got its version from and I don't see its sources. 

Our dftd3 references then again the old version from dftb+, that is now deprecated and not updated anymore in favour of simple-dftd3, which is is not supported by Psi4 and QCEngine.

sheepforce:
@markuskowa any opinion on what to do best? I think we have three options, all of them bad:

  * Remove dft-d support from Psi4
  * Package the dft-d3 binaries from CondaForge
  * Try the unstable DFT-D3 link from the Mulliken center and hope that this somehow works as 3.2.1

markuskowa:
Inspecting the Psi4 tree, I found the conda recipe (https://github.com/psi4/psi4meta/tree/6f6493fbdf14de0b4dc1c3e554d67bc662167b31/conda-recipes/dftd3), which points to https://github.com/loriab/dftd3. There is a version 3.2.1. Could that be it?

sheepforce:
Good spot, that should be fixing it.

markuskowa:
Psi4 works now, but pysisyphus now fails a bunch of tests.

sheepforce:
> Psi4 works now, but pysisyphus now fails a bunch of tests.

Argh, pysisyphus is really our canary bird ... :expressionless:  :sweat_smile:  I will have a look what is failing and see if I can fix it.

sheepforce:
It is the upstream PySCF update that slightly changed results. The failing test cases are mostly HF calculations on small molecules with small basis sets, all of them using PySCF. The PySCF test suite passes everything and the Pysisyphus results are slightly off, say 5e-4 instead of the allowed 2e-4. This can be reproduced in a plain pyenv with PySCF outside of NixOS as well. Johannes will probably adapt the tests at some point, but for now we can think about disabling the specific ones. What do you think, @markuskowa ?

markuskowa:
>Johannes will probably adapt the tests at some point, but for now we can think about disabling the specific ones. What do you think, @markuskowa ?

Yes, I think that is a good solution for now. The test criteria are a little too tight anyway?

PR #281, opened 2023-05-05
sharc: 2.1.1 -> 3.0.1

sheepforce:
Recently, SHARC updated to Python3 and can be unbroken, finally.

Currently we cannot build PySharc as it requires HDF4 and HDF4 in Nixpkgs has explicitly disabled Fortran support. If enable Fortran bindings in HDF4 upstream, we can try to get Pysharc running again.

cc @ziechys


pkgs/apps/sharc/default.nix
```diff
@@ -112,6 +141,6 @@ in stdenv.mkDerivation {
     license = licenses.gpl3;
     maintainers = [ maintainers.markuskowa ];
     platforms = platforms.linux;
-    broken = true; # numpy does not support python-2.7 anymore
+    # broken = true; # numpy does not support python-2.7 anymore
```
markuskowa, line 144:
```suggestion
```

markuskowa (commented):
> Currently we cannot build PySharc as it requires HDF4 and HDF4 in Nixpkgs has explicitly disabled Fortran support. If enable Fortran bindings in HDF4 upstream, we can try to get Pysharc running again.

It's bit ironic that they just got rid of an outdated Python version, but now introduce a dependency to an outdated file format. :expressionless:
I guess we can just merge it without HDF4 support for now an enable it later when hdf4 is fixed upstream.

PR #288, opened 2023-05-26
psi4: 1.7 -> 1.8

sheepforce:
Here is another, quite interesting Psi4 update: https://github.com/psi4/psi4/releases/tag/v1.8
Among other things, we get TD-DFT for GGA :partying_face: 

However, no Psi4 release without a complete overhaul of how to handle libint ... While it now changed to standard basis function ordering, they maintain a fork with support for Boost and CMake, which is not available in upstream libint :smiling_face_with_tear: Consequently, I am now falling back to the option again to build Psi4's libint version together with Psi4, which, of course, takes forever.


markuskowa:
>  now falling back to the option again to build Psi4's libint version together with Psi4, which, of course, takes forever.

It was actually pretty quick (pysisiphys actually took much longer).

PR #306, opened 2023-08-02
flake: disable gaussview in packages

sheepforce:
Gaussview is not disabled by default as the other unfree packages in the exported flake outputs, resulting in Hydra failures and a formal inconsistency. Have disabled it, too.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -25,6 +25,7 @@
               cefine = null;
               cfour = null;
               gamess-us = null;
+              gaussview = null;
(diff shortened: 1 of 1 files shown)


markuskowa (approved):
I have not encountered any Hydra problems though. I guess it depends on the setting, if you already have it in your store?

sheepforce:
Yeah it just complains that it fails to build due to missing sources as any other of the `requireFile` packages would fail.

PR #307, opened 2023-08-02
psi4: fix compilation with hdf5 1.14.1-2, reenable adcc

### LINKED ISSUE #303: psi4: broken on nixpkgs-unstable lastest
```
-- Suitable Libint2 could not be located, Building Libint2 from generator source instead.
CMake Warning at external/upstream/libint2/CMakeLists.txt:115 (message):
  Libint2 generator source build requires Ninja (not Makefile) builder
  and Boost, Eigen, and GMPXX dependencies.  The build will be very long for
  non-trivial AM, and parallelism won't help for the generation phase.  Edit
  external/upstream/libint2/CMakeLists.txt for AM, derivative, and integral
  classes settings.
```

sheepforce:
Closes #303

### DIFF
--- a/pkgs/apps/adcc/default.nix
+++ b/pkgs/apps/adcc/default.nix
@@ -9,13 +9,13 @@ assert !blas.isILP64;
 
 buildPythonPackage rec {
   pname = "adcc";
-  version = "0.15.14";
+  version = "0.15.16";
 
   src = fetchFromGitHub {
     owner = "adc-connect";
     repo = "adcc";
     rev = "v${version}";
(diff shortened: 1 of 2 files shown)


sheepforce:
Should work now. What have I thought; updating the commit of libint in a stable release branch? Of course this would break Psi4 :roll_eyes:
We are back at the original commit, now.
However, upstream's Psi4  CMakeList.txt is broken itself and passing `"-DBUILD_Libint2_GENERATOR=ON"` would not work with the original Psi4 version. It only works in our case because of the pinning. Meh! :expressionless:

sheepforce:
Indeed, works: https://nix-qchem-hydra.fysik.su.se//eval/229490 :partying_face:

PR #312, opened 2023-08-08
Xtb iff

sheepforce:
Enables crest to perform quantum cluster growth calculations with the XTB iff

### DIFF
--- a/pkgs/apps/xtb-iff/default.nix
+++ b/pkgs/apps/xtb-iff/default.nix
@@ -1,0 +1,31 @@
+{ stdenv, fetchurl, autoPatchelfHook }:
+
+stdenv.mkDerivation rec {
+  pname = "xtb-iff";
+  version = "1.1";
+
+  src = fetchurl {
+    url = "https://github.com/grimme-lab/xtbiff/releases/download/v${version}/xtbiff.tar.xz";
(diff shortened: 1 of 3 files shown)


pkgs/apps/xtb-iff/default.nix
```diff
@@ -0,0 +1,31 @@
...
+
+  installPhase = ''
+    runHook preInstall
+
+    mkdir -p $out/bin
+    install xtbiff $out/bin
+
+    runHook postInstall
+  '';
+}
```
markuskowa, line 31: A meta field would be nice here.

pkgs/apps/xtb-iff/default.nix
```diff
@@ -0,0 +1,31 @@
...
+
+  src = fetchurl {
+    url = "https://github.com/grimme-lab/xtbiff/releases/download/v${version}/xtbiff.tar.xz";
+    hash = "sha256-bV3uELo57LHiY3S3v4LiNk2OvtW281zc5M7XMkXClvU=";
+  };
+
+  unpackPhase = ''
+    runHook preUnpack
+
+    tar -xvf ${src}
```
markuskowa, line 15: `stdenv` can not handle xz compression?

sheepforce: It can but it will complain that no source directory is produced. It is really just a single, statically linked executable in tar archive ... For some reasons ...

Issue #313, opened 2023-08-09
Dalton fails to build


sheepforce:
From our Hydra:
```
exporting https://gitlab.com/dalton/dalton.git (rev 9d7c5e435b75a9695d5ac8714121d12e6486149f) into /nix/store/kbga2mxlp180w21gck2rkd0nm22ygrpd-dalton-9d7c5e4
Initialized empty Git repository in /nix/store/kbga2mxlp180w21gck2rkd0nm22ygrpd-dalton-9d7c5e4/.git/
remote: Enumerating objects: 4666, done.
remote: Counting objects: 100% (4666/4666), done.
remote: Compressing objects: 100% (3460/3460), done.
remote: Total 4666 (delta 1539), reused 2980 (delta 1114), pack-reused 0
Receiving objects: 100% (4666/4666), 26.87 MiB | 8.48 MiB/s, done.
Resolving deltas: 100% (1539/1539), done.
From https://gitlab.com/dalton/dalton
 * branch            9d7c5e435b75a9695d5ac8714121d12e6486149f -> FETCH_HEAD
Updating files: 100% (4425/4425), done.
Switched to a new branch 'fetchgit'
Submodule 'external/gen1int' (https://gitlab.com/bingao/gen1int.git) registered for path 'external/gen1int'
Submodule 'external/pelib' (https://gitlab.com/pe-software/pelib.git) registered for path 'external/pelib'
Submodule 'external/qfitlib' (https://github.com/cstein/qfitlib.git) registered for path 'external/qfitlib'
Cloning into '/nix/store/kbga2mxlp180w21gck2rkd0nm22ygrpd-dalton-9d7c5e4/external/gen1int'...
Cloning into '/nix/store/kbga2mxlp180w21gck2rkd0nm22ygrpd-dalton-9d7c5e4/external/qfitlib'...
Cloning into '/nix/store/kbga2mxlp180w21gck2rkd0nm22ygrpd-dalton-9d7c5e4/external/pelib'...
Submodule path 'external/gen1int': checked out '1e4148ecd676761b3399801acba443925a1fee6b'
Submodule path 'external/pelib': checked out '19dd0e91afbd18b7c9f2611b9978b1aeaa1c19f9'
Submodule path 'external/qfitlib': checked out '1acdc9863fdeae2cdbc7f5a599413257a095b8ad'
fatal: not a git repository (or any parent up to mount point /)
[...]

PR #315, opened 2023-08-15
dalton: take care of git submodules manually

sheepforce:
Proposed fix for #313 

All submodules are handled by nixpkgs fetchers individually and pinned to the commits used in the 2020.1 release tag. This works but is unpleasantly verbose.


markuskowa:
> This works but is unpleasantly verbose.

Agree, but I also have no better idea at the moment. This is a reoccurring pattern with a solution that I have also used for cases where the build system calls `git clone`.

PR #320, opened 2023-08-16
xtb-python: fix meson.build

sheepforce:
From the xtb-python readme:

```
This repository hosts the Python API for the extended tight binding (xtb) program.

    Warning ⚠️ : xtb-python is no longer in active development. We recommend using [tblite](https://github.com/tblite/tblite) instead.

The idea of this project is to provide the xtb API for Python without requiring an additional xtb installation.
```

If it fails in the future and fixing becomes too complex I think we should just mark it as broken.


markuskowa:
> If it fails in the future and fixing becomes too complex I think we should just mark it as broken.

...and eventually remove it if that feature it is not supported anymore by the developers.

PR #323, opened 2023-08-18
overlay: fix system impurity

sheepforce:
The pinned overlay `default.nix` exposed via the flake as `overlays.qchem'` throws an error in a pure flake evaluation as we did not specify the `system` architecture in `default.nix`. This should be fixed by this PR.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -11,7 +11,7 @@ let
 
   pkgs = nixpkgs {
     overlays = [ (import ./overlay.nix) ];
-    inherit (prev) config;
+    inherit (prev) config system;
   };
 
 in


default.nix
```diff
@@ -11,6 +11,7 @@ let
 
   pkgs = nixpkgs {
     overlays = [ (import ./overlay.nix) ];
+    system = "x86_64-linux";
     inherit (prev) config;
```
markuskowa, line 15:
```suggestion
    inherit (prev) config system;
```
That should be cleaner?

sheepforce: Ah of course, even better!

PR #366, opened 2023-10-24
CUDA support: turn on CUDA support in nixpkgs via supportCuda

markuskowa:
This is much easier than dealing with a seperate overlay.

CUDA support flags have been unified in https://github.com/NixOS/nixpkgs/pull/224068, so we can just use it.


sheepforce:
LGTM! :sheep: Nice to have this unified upstream.

sheepforce (approved):

PR #368, opened 2023-10-27
sgroup: init at 1.0, add as exciting addon

sheepforce:
This PR provides the sgroup utility, used by exciting to determine symmetries. Being a download link from the SI of an Elsevier paper is an interesting step in software distribution ... :)

@mathunje please let me know if it works as expected before we merge it.

### DIFF
--- a/pkgs/apps/exciting/default.nix
+++ b/pkgs/apps/exciting/default.nix
@@ -1,6 +1,17 @@
-{ lib, stdenv, fetchFromGitHub
-, gfortran, perl, libxslt, runtimeShell
-, blas, lapack, mpi, arpack }:
+{ lib
+, stdenv
+, fetchFromGitHub
+, gfortran
+, perl
+, libxslt
+, runtimeShell
+, blas
+, lapack
+, mpi
+, arpack
+, enableSgroup ? false
+, sgroup
+}:
 
 stdenv.mkDerivation rec {
   pname = "exciting";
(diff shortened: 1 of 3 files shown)


pkgs/apps/exciting/default.nix
```diff
@@ -1,6 +1,6 @@
 { lib, stdenv, fetchFromGitHub
 , gfortran, perl, libxslt, runtimeShell
-, blas, lapack, mpi, arpack }:
+, blas, lapack, mpi, arpack, sgroup }:
```
markuskowa, line 3: I wonder if this should be optional? It turns exciting into a package that needs  unfree software to be built.

sheepforce: I can make it optional, yes. I've kept "unfree" in sgroup to be safe. In fact, they do not specify any license or copyright. :grimacing:

markuskowa: On the `homepage` they refer to the [CPC license](https://beta.elsevier.com/about/policies-and-standards/open-access-licenses/elsevier-user/cpc), which reads like "unfree" to me. I think the classification as unfree is appropriate here.

Cite from the CPC license:
1. This licence entitles the licensee (one person) and the licensee's research group to obtain a copy of the source or executable code and to use the acquired program for academic or non-profit use within a research group; or, it entitles the licensee (one company, organisation or computing centre) to install the program and allow access to the executable code to members of the licensee's organisation for academic or non-profit use. **No user or site will re-distribute the source code or executable code to a third party in original or modified form without the written permission of the author**.

markuskowa (approved):

PR #370, opened 2023-10-30
sharc: split outputs, split binaries and wrappers.

markuskowa:
Optimize for cache space and build time:

* split up into sharc-unwrapped and the actual wrappers.
  This saves a lot of space and some build time. We need to build the SHARC binaries only once.
  Space saving: N * 5 MB vs. 5 MB + N * 0.5 MB

* use separate outputs for the documentation and the tests. Save space in run time closure.
  binaries (5 MB), manual (7.5 MB), tests (30 MB).

Overall savings with both changes: N * 43 MB vs. 5 MB + 30 MB + 7.5 MB + N * 0.5 MB

### DIFF
--- a/overlay.nix
+++ b/overlay.nix
@@ -254,12 +254,15 @@ let
 
         sgroup = callPackage ./pkgs/apps/sgroup { };
 
-        # blank version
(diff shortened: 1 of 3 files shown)


sheepforce (approved):
LGTM

PR #371, opened 2023-11-03
exciting: propagate MPI

sheepforce:
Exciting is missing the propagation of MPI for runtime usage. Has been added.


pkgs/apps/exciting/default.nix
```diff
@@ -74,7 +74,10 @@ stdenv.mkDerivation rec {
     mpi
```
markuskowa, line 74: This is redundant now and can be removed.

PR #381, opened 2023-12-11
Psi4: 1.8.2 -> 1.9

sheepforce:
https://github.com/psi4/psi4/releases/tag/v1.9

PyLibefp now also requires a special commit to work and needs to be built with Psi4's Cmake to be recognised. A normally packaged version is not working ...

### DIFF
--- a/pkgs/apps/psi4/default.nix
+++ b/pkgs/apps/psi4/default.nix
@@ -16,7 +16,6 @@
 , qcelemental
 , qcengine
 , numpy
-, pylibefp
 , deepdiff
 , blas
 , lapack
@@ -25,7 +24,6 @@
 , dkh
 , dftd3
 , pcmsolver
-, libefp
 , libecpint
 , cppe
 , chemps2
@@ -135,7 +133,7 @@ let
 in
 buildPythonPackage rec {
   pname = "psi4";
-  version = "1.8.2";
+  version = "1.9";
 
   nativeBuildInputs = [
     cmake
@@ -153,7 +151,6 @@ buildPythonPackage rec {
     lapack
     dkh_
     pcmsolver_
-    libefp
     libecpint
     cppe
     chemps2_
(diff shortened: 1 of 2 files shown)


sheepforce:
Had to disable libEFP as I couldn't get Psi4's cmake to find it after building it itself internally :o

pkgs/apps/psi4/default.nix
```diff
@@ -16,7 +16,6 @@
 , qcelemental
 , qcengine
 , numpy
-, pylibefp
```
markuskowa, line 19: If it is removed. Is the upgrade below still needed?

pkgs/apps/psi4/default.nix
```diff
@@ -211,7 +207,7 @@ buildPythonPackage rec {
     "-DCMAKE_INSIST_FIND_PACKAGE_gau2grid=ON"
     "-Dgau2grid_DIR=${gau2grid}/share/cmake/gau2grid"
     # libint
-    "-DMAX_AM_ERI=6"
+    "-DMAX_AM_ERI=4"
```
markuskowa, line 210: Any special reason reduce the MAX_AM?

pkgs/apps/psi4/default.nix
sheepforce: Not necessarily. 0.6.1 is not the latest release, that is 0.6.2. Should I upgrade go to this then?

markuskowa: Yes, a release is always cleaner than upgrading to unstable (unless it is really necessary).

pkgs/apps/psi4/default.nix
sheepforce: Ahh this was a leftover from playing around to get build time down from 2 hours :grimacing: Back at 6.

markuskowa: On the big AMD machine with 64 threads it takes about an hour.

PR #397, opened 2024-02-15
Crest updates

sheepforce:
Updates crest to current head of the main branch. This is mainly intended to mitigate https://github.com/crest-lab/crest/issues/216 (breaking QCG with gfortran). The current solution fake-initialises gfn0 and gfnff as submodules as they can not be found by CMake as external packages for some reasons.

### DIFF
--- a/pkgs/apps/crest/default.nix
+++ b/pkgs/apps/crest/default.nix
@@ -1,45 +1,53 @@
 { stdenv
 , lib
-, fetchpatch
-, makeWrapper
 , cmake
 , gfortran
 , blas
 , lapack
 , fetchFromGitHub
-, xtb
-, xtb-iff
+, tblite
+, mctc-lib
+, toml-f
+, simple-dftd3
+, dftd4
+, multicharge
+, gfn0
+, gfnff
 }:
 
 stdenv.mkDerivation rec {
   pname = "crest";
-  version = "2.12";
+  version = "unstable-2024-02-14";
 
   src = fetchFromGitHub {
     owner = "grimme-lab";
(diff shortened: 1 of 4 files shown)


pkgs/apps/crest/default.nix
```diff
@@ -1,45 +1,53 @@
...
+, dftd4
+, multicharge
+, gfn0
+, gfnff
 }:
 
 stdenv.mkDerivation rec {
   pname = "crest";
-  version = "2.12";
+  version = "unstable-14-02-2024";
```
markuskowa, line 20:
```suggestion
  version = "unstable-2024-02-14";
```

pkgs/apps/gfn0/default.nix
```diff
@@ -0,0 +1,43 @@
...
+  ];
+
+  buildInputs = [
+    blas
+    lapack
+  ];
+
+  postInstall = ''
+    substituteInPlace $out/lib/cmake/${pname}/${pname}-*.cmake \
+      --replace "libgfn0.a" "libgfn0.${stdenv.hostPlatform.extensions.sharedLibrary}"
```
markuskowa, line 33: This replaces a static library name with a shared one. Does this yield the correct result?

sheepforce: Yes, the Cmake file hardcodes the `.a` suffix which is not true if we build dynamically as given by the `cmakeFlags`.

sheepforce: Ah right, it should be conditional ... :grimacing:

PR #413, opened 2024-03-15
salmon: init at 2.2.0

sheepforce:
Adds the Salmon TDDFT code https://salmon-tddft.jp/

The build system is terribly broken and I had to patch out overrides of fortran intrinsics , the entire Scalapack detection and all link libraries need to be specified manually. The upstream instructs me to call a `configure.py` wrapper around CMake, but this makes things worse. This should be the fully featured and working version with some terrible hacks around the broken CMake.

### DIFF
--- a/overlay.nix
+++ b/overlay.nix
@@ -249,6 +249,8 @@ let
 
         qmcpack = callPackage ./pkgs/apps/qmcpack { };
 
+        salmon = callPackage ./pkgs/apps/salmon { };
+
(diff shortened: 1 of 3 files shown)


sheepforce:
I've tried to get the test suite running and some test pass. However, many other require network access and wget to download parameter files, other on the fly during test execution create custom shell wrappers calling MPI and hardcode `/bin/sh` as interpreter into those scripts.

PR #415, opened 2024-03-21
turbomole: 7.7 -> 7.8.1

sheepforce:
I've got the recent Turbomole version from the developers and here is the update. Most interestingly now including RT-TDDFET. Turbomole now comes as a self-extracting archive, which changed how I'm unpacking it in Nix.

I've run nix fmt somewhere in the middle of development on the file, so the diff is unfortunately more verbose than it needed to be, sorry :grimacing:

### DIFF
--- a/pkgs/apps/turbomole/default.nix
+++ b/pkgs/apps/turbomole/default.nix
@@ -1,15 +1,30 @@
-{ stdenv, lib, makeWrapper, requireFile, autoPatchelfHook, writeTextFile, perl, sysstat, zlib,
-  glibc, gcc-unwrapped, which, less, more, coreutils,
-  # Configuration
+{ stdenv
+, lib
+, makeWrapper
+, requireFile
+, autoPatchelfHook
+, writeTextFile
+, perl
+, sysstat
+, zlib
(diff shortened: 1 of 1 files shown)


sheepforce:
Runs perfectly locally, all modules seem to work.

pkgs/apps/turbomole/default.nix
```diff
@@ -30,31 +45,40 @@ in stdenv.mkDerivation rec {
   ];
 
   src = requireFile {
-    sha256 = "b903ffb4fb9d1a1dbca81187f039df99c9fc22a4eefce5566db852da4991795d";
-    name = "turbolinux${lib.replaceStrings ["."] [""] version}_TMG.tar.gz";
+    sha256 = "21518913591e7ab58b3af1c126fdc37fa31cb5507d523b4b04fbe57f63916b5e";
+    name = "turbolinux${lib.replaceStrings ["."] [""] version}-TMG.bin";
```
markuskowa, line 49: The download is offered as .zip file not directly as the .bin. I think we should ingest the zip file here.

PR #426, opened 2024-05-03
mdtraj: move cython to nativeBuildInputs

mcwitt:
This fixes an issue I'm having using the overlay with recent nixpkgs versions, where `cython` must be overriden to `cython_0` (since mdtraj needs 0.x), and this causes a collision in the environment when other packages propagate a different cython version.

My understanding is that because cython is [declared as a build-system dependency](https://github.com/mdtraj/mdtraj/blob/43d8194bbc2e7f97a74b69fa02708bf17589529e/pyproject.toml#L7), it's more appropriate to put it in `nativeBuildInputs` (and this also fixes the collision issue).

### DIFF
--- a/pkgs/lib/mdtraj/default.nix
+++ b/pkgs/lib/mdtraj/default.nix
@@ -31,9 +31,12 @@ buildPythonPackage rec {
 
   buildInputs = [ zlib ];
 
+  nativeBuildInputs = [
+    cython
+  ];
+
   propagatedBuildInputs = [
     setuptools
-    cython
     numpy
     pyparsing
     astunparse


mcwitt:
I'd missed that mdtraj is now included in nixpkgs as of https://github.com/NixOS/nixpkgs/commit/71f5ffa84e539e420a35921bec2889d748c58cc7

The nixpkgs version does put `cython` in `nativeBuildInputs`, but I have another issue with it: 

```
nix shell --refresh --impure --expr 'let nixpkgs = builtins.getFlake "github:nixos/nixpkgs/nixpkgs-unstable"; pkgs = nixpkgs.legacyPackages.x86_64-linux; in (pkgs.python311.withPackages (ps: [ ps.mdtraj ]))' --command python -c "import mdtraj"
```
produces ModuleNotFoundError: No module named 'pkg_resources'. It seems like this might be due to the nixpkgs version including `setuptools` in `nativeBuildInputs`.

markuskowa:
We should actually remove `mdtraj` from NixOS-QChem in favor of the upstream package (looks like we forgot to do that). 
That should then also solve your problem?

mcwitt:
> We should actually remove `mdtraj` from NixOS-QChem in favor of the upstream package (looks like we forgot to do that). That should then also solve your problem?

That works for me, thanks! (I'm having a separate issue with the upstream package, but have opened https://github.com/NixOS/nixpkgs/issues/308847 for that.)

mcwitt:
Closed in favor of #427

PR #429, opened 2024-05-07
crest: 3.0 -> 3.0.1

sheepforce:
Bugfix release for crest: https://github.com/crest-lab/crest/releases/tag/v3.0.1

Unfortunately, crest now also requires a more recent tblite version than the last released one. This was already required for xTB and is not so nice for a bugfix release. We could think about overwriting tblite either here or even upstream to use an unstable version.


overlay.nix
```diff
@@ -135,7 +135,17 @@ let
 
         cpcm-x = callPackage ./pkgs/lib/cpcm-x { };
 
-        crest = callPackage ./pkgs/apps/crest { };
+        crest = callPackage ./pkgs/apps/crest {
+          tblite = super.tblite.overrideAttrs (old: {
```
markuskowa, line 139: I think that is ok for now. Maybe add a short comment here about the purpose and when this can be removed?

PR #432, opened 2024-05-13
WiP: gamess-us: 2021R2P1 -> 2023R1P1

sheepforce:
Finally updates GAMESS to the latest release. Again, unfortunately older releases are not available for download anymore and have to be archived manually.

I've removed the BLAS choices between OpenBLAS and MKL. I've never used it with MKL anyway, OpenBLAS is more portable with fewer surprises and then one could/should also support blis/libflame and so on. 

Also invocation of the rungms script has changed. Now the third argument `NCPUS` must either be 1 (to use 1 "Processor" which in fact is a compute node. For values > 1 it must instead be a MPI-style host file unless submitted via Torque or SGE ...?) and the fourth argument `PPN` "Processors" must be the name of cores per node. Whatever ... Tests had to be adapted ...


markuskowa:
> I've removed the BLAS choices between OpenBLAS and MKL. I've never used it with MKL anyway, OpenBLAS is more portable with fewer surprises and then one could/should also support blis/libflame and so on.

I am fine with that. It is a complicated package to handle. This reduces the maintenance burden, which is already high enough in this case.

sheepforce:
Works as pushed.

PR #438, opened 2024-05-20
Update flake.lock/package list

markuskowa:
qchem.realease-barrier has passed

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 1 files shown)


markuskowa:
@sheepforce Note that the rebuild workflow is now broken, since nix can not find the AmberTools sources. I added them to cachix but that does not seem to help here.

markuskowa:
Note: the .drv file of the sources needs also to be in the cache. Now it works and the workflow can build AmberTools.

PR #455, opened 2024-06-03
Convert directory structure to pkgs/by-name

markuskowa:
This an attempt to imitate what upstream has done with RFC 140.
The advantage is a much more slim overlay.nix and less boilerplate code.

We can also do this for all Python packages by moving them into their own `pkgs/python-by-name/` sub-tree.


sheepforce:
In principle I like this change. I guess overriding a dependency then requires moving something back to the `pkgs/{apps,libs}` structure? Or is there any way to override a package that is called by name? I also didn't understand yet how this is supposed to happen in nixpkgs.

sheepforce:
> I am bit confused why flake check does get the 'by-name' path into its store.

I'm afraid I don't understand this. What do you mean? :)

markuskowa:
Now also with `pkgs/python-by-name`.

overlay.nix
```diff
@@ -49,7 +31,9 @@ let
 
     in
     {
-      "${subset}" = optUpstream // {
+      "${subset}" = optUpstream
+        // (qlib.pkgs-by-name callPackage ./pkgs/by-name)
```
sheepforce, line 35: Should this be `lib.attrsets.recursiveUpdate` to avoid some surprises maybe?

sheepforce (commented):
I saw that also packages that require an override are now in `by-name`, like `crest`. Others are now, e.g. chemps2. Is there a reason for that or is it simply not finished yet?

sheepforce:
The `--no-build`  issue with the flake honestly sounds like a bug with the lazy filesystem tree evaluation in recentish Nix versions. Maybe this is worth an upstream bug?

overlay.nix
markuskowa: I was thinking about that. The attribute sets are relatively flat, with the exception of python that that gets handled separately. I would rather expect to get surprises with `recursiveUpdate`.

markuskowa:
> I saw that also packages that require an override are now in `by-name`, like `crest`. Others are now, e.g. chemps2. Is there a reason for that or is it simply not finished yet?

Still work in progress, but I except a very few oddballs to be left over that  will not fit the scheme.

> The --no-build issue with the flake honestly sounds like a bug with the lazy filesystem tree evaluation in recentish Nix versions. Maybe this is worth an upstream issue?

Yes, could be a bug, but I was also wondering if I am just using it wrong. A similar thing happens when you run the updater script for the package list files. This will also fail, if one does not call nix-instantiate of nix-build before running the script.

markuskowa:
I think now I have everything. Everything should either go into `pkgs/by-name` or `pkgs/python-by-name` unless
* it does not return a derivation (e.g. returns a function).
* it is called by `callPackage` from a different scope (e.g. Qt5).
* it is packaged by more than one .nix file, e.g. one for a C build and one for a python, or wrapped/unwrapped packages.
* it requires a license file or opt path from config (e.g. molpro, q-chem)

All packages left in `pkgs/apps` fall under one of these 4 cases.

overlay.nix
sheepforce: Hu, could also be, yes :sweat_smile: The language subsets are strange, same with Haskell. Then let's see if this works as expected or someone complains later.

sheepforce (approved):

markuskowa:
Hydra shows all green. We did not loose any packages. Are we brave to merge it see what happens?

PR #466, opened 2024-06-11
lib: fix by-name path construction

markuskowa:
Using a + instead of string interpolation allows to find paths in the directory tree instead of /nix/store. This is required for use  with nix-update


sheepforce:
Crazy, how did you find out? Seems very subtle.

markuskowa:
> Crazy, how did you find out? Seems very subtle.

Yes, it is crazy subtle. I was trying nix-update, which did was trying to locate the package file in the nix store (and then taking a look at the nixpkgs implementation of by-name).

PR #514, opened 2024-08-06
WIP: Flake lock update

markuskowa:
See https://github.com/Nix-QChem/NixOS-QChem/issues/513

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1720181791,
-        "narHash": "sha256-i4vJL12/AdyuQuviMMd1Hk2tsGt02hDNhA0Zj1m16N8=",
+        "lastModified": 1725194671,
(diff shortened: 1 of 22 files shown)


sheepforce:
I hope the hydra evaluation errors are gone now. I currently have a hydra job running for everything, but locally everything was working now, again.

sheepforce:
If everything turns out to be working, we could possibly squash some commits in this branch, to clear the iterations between broken / non-broken, python3 / python311 and multiple flake.lock updates?

sheepforce:
On my Hydra two packages still fail: `amd-scalapack` with test timeouts and `pysisyphus` with one failing test, which is caused by a minor change in ASE. Best case scenario is a quick upstream fix for the pysisyphus issue. For Scalapack I've no idea.

sheepforce:
>  Would be good if could also fix pysisphus before merging.

OK, so hopefully this is also done now. Do we have anything else left, except adm-scalapack timing out? :)

PR #534, opened 2024-09-16
orca: add otool_external hook for external optimizer

markuskowa:
As described in section 9.24.6 in the ORCA manual, ORCA can use an external program for geometry optimization.
However, this script needs to be right next to the ORCA executables. 

We provide two options now:
1. A simple wrapper that forwards the command to a script set in $ORCA_OTOOL_EXTERNAL (default)
2. Fully declarative: provide a path to a script via inputs (requires orca rebuild).

### DIFF
--- a/pkgs/by-name/orca/package.nix
+++ b/pkgs/by-name/orca/package.nix
@@ -15,12 +15,11 @@ stdenv.mkDerivation {
 
   src =
     if enableAvx2 then
-      requireFile
-        {
+      requireFile {
           name = "orca_6_0_0_linux_x86-64_avx2_shared_openmpi416.tar.xz";
           sha256 = "sha256-AsISlO/nsbch4my5D5juFa1oLQKAcgG30hff5nkFov0=";
           url = "https://orcaforum.kofo.mpg.de/app.php/portal";
-        } else
+      } else
       requireFile {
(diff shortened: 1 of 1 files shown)


markuskowa:
> Hm, I can only find these options in chapter 7.26.6. There is not even chapter 9 in the ORCA manual, no?
> https://www.faccts.de/docs/orca/6.0/manual/contents/detailed/geomopt.html#orca-as-external-optimizer

Oops old version 5..., but yes.

markuskowa:
I removed the wrapper script. This may have made sense for orca-5, but since you can specify it in the input or via $EXTOPTEXE in orca-6, this would be redundant.

pkgs/by-name/orca/package.nix
```diff
@@ -7,6 +7,10 @@
 , openssh
 , xtb
 , enableAvx2 ? true
+
+# Provide path to a custom script for otool_external
+, withCustomOtoolExternal ? false
+, customOtoolExternal ? null
```
sheepforce, line 13: Should there maybe be an assert that `customOtoolExternal` is not null, if `withCustomOtoolExternal` is set?

markuskowa: ok, now it should fail gracefully.

sheepforce (approved):

PR #537, opened 2024-09-17
gamess-us: 2023R1P1 -> 2024R2P1

sheepforce:
Updates GAMESS to the current version: https://www.msg.chem.iastate.edu/gamess/versions.html

EDIT: Somewhere in the refactoring of OpenMPI upstream, `orterun` was lost from `mpi.bin`. It is merely an alias to `mpiexec`. However, `rungms` wants to use `orterun`. 

  * [x] Patch rungms to use `mpiexec` instead of `orterun`.

### DIFF
--- a/pkgs/by-name/gamess-us/gcc-config.patch
+++ b/pkgs/by-name/gamess-us/gcc-config.patch
@@ -1,40 +1,34 @@
 diff --git a/comp b/comp
-index 3cdf0a9..0fc9da2 100755
+index 52602ff..1cb96ac 100755
 --- a/comp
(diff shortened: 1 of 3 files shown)


markuskowa:
> Somewhere in the refactoring of OpenMPI upstream, orterun was lost from mpi.bin

OpenMPI-5 has switched to from the ORTE runtime env to PRRTE, thus not orterun anymore. It's a weird choice anyway to access orterun instead of mpirun/mpiexec, as these are standardized.

Otherwise, the diff looks harmless this time :)

PR #538, opened 2024-09-17
dice: init at unstable-2024-07-02

sheepforce:
Adds the Dice program for Heatbath CI. Can be used for large active space CASSCF in PySCF and Molcas. I've also added a version of Molcas that builds with DICE support. However, turns out there are some exclusive things going on in Molcas.

  * When buiding with `-DDMRG=ON` and `-DNEVPT2=ON` (upstream nixpkgs), you can't have `-DDICE=ON`. However, the EXPBAS module is available
  * When enabling LibWFA Support (`molcas` attribute in nixpkgs), EXPBAS support is disabled due to required `-DH5_USE_110_API` flag
  * I've added a version `molcasDice`, that disables DMRG and NEVPT2, but has DICE and LibWFA support, but not EXPBAS support. 

Conclusion; Molcas build system is sensitive and has some mutually exclusive options. Maybe it would be nice to have these things configurable, but that is not that easy with upstream molcas but our DICE and LibWFA builds.

### DIFF
--- a/overlay.nix
+++ b/overlay.nix
@@ -132,6 +132,10 @@ let
 
         dftbplus = super.python3.pkgs.toPythonApplication self.python3.pkgs.dftbplus;
 
+        dice = callPackage ./pkgs/by-name/dice/package.nix {
+          boost = self.boost-mpi;
+        };
+
         dirac = callPackage ./pkgs/by-name/dirac/package.nix {
           inherit (self) exatensor;
         };
--- a/overlay.nix
+++ b/overlay.nix
@@ -148,6 +148,8 @@ let
 
         iboview = prev.libsForQt5.callPackage ./pkgs/apps/iboview { };
 
+        # Molcas with optimisation and LibWFA support. Note, that this disables
+        # EXPBAS support, however.
         molcas = let
           molcasOpt = prev.openmolcas.override {
             stdenv = aggressiveStdenv;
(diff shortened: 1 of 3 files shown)


overlay.nix
```diff
@@ -132,6 +132,10 @@ let
 
         dftbplus = super.python3.pkgs.toPythonApplication self.python3.pkgs.dftbplus;
 
+        dice = callPackage ./pkgs/by-name/dice/package.nix {
```
markuskowa, line 135: You need to rebuild boost with an optimized MPI, otherwise you depend on two different version (optimized and unoptimized).


```
diff --git a/nixpkgs-opt.nix b/nixpkgs-opt.nix
index 3db064f..389b865 100644
--- a/nixpkgs-opt.nix
+++ b/nixpkgs-opt.nix
@@ -18,6 +18,7 @@ let
     avogadro2 = recallPackage avogadro2 {};
     arpack = recallPackage arpack {};
     arpack-mpi = recallPackage arpack { useMpi = true; };
+    boost-mpi = recallPackage boost { useMpi = true; mpi = self.mpi; };
     cp2k = recallPackage cp2k {};
     fftw = recallPackage fftw {};
     dkh = recallPackage dkh {};
diff --git a/overlay.nix b/overlay.nix
index b63c9d4..614dd2c 100644
--- a/overlay.nix
+++ b/overlay.nix
@@ -133,7 +133,7 @@ let
         dftbplus = super.python3.pkgs.toPythonApplication self.python3.pkgs.dftbplus;
 
         dice = callPackage ./pkgs/by-name/dice/package.nix {
-          boost = final.boost.override { useMpi = true; };
+          boost = self.boost-mpi;
         };
 
         dirac = callPackage ./pkgs/by-name/dirac/package.nix {
```

pkgs/by-name/dice/package.nix
```diff
@@ -0,0 +1,64 @@
...
+, eigen
+, blas
+, lapack
+, mpi
+}:
+
+
+stdenv.mkDerivation {
+  pname = "dice";
+  version = "unstable-2024-07-02";
```
markuskowa, line 15:
```suggestion
  version = "2024-07-02";
```

overlay.nix
sheepforce: Oh I wasn't aware that we have a recalled MPI. Good to know :sheep:

markuskowa: This is a bit hidden unfortunately and a result of the optimized upstream package set.

markuskowa (approved):

Issue #540, opened 2024-09-18
more python upgrade trouble: pycolt and theodore


markuskowa:
Fails now in the evaluation phase with:
```
error: pytest-runner has been removed as it uses deprecated features of setuptools and is deprecated by upstream
```

CC @sheepforce

sheepforce:
Given the package is somehow build in a standard way (a big "if", I know), [the correct solution is the `pythonRelaxDepsHook`](https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/python.section.md#using-pythonrelaxdepshook-using-pythonrelaxdepshook) and the `removeDeps` list. For `setup.py` this should work.

sheepforce:
Closed via #541

PR #541, opened 2024-09-18
Pytest runner

sheepforce:
Removes `pytest-runner` dependencies from Theodore and PyColt. Dirty workaround for #540


markuskowa:
Thanks! That is better than my solution.

PR #549, opened 2024-09-25
qmcpack: add utility scripts and python modules

### LINKED ISSUE #547: QMCPack Utilities
The QMCPack repo offers a lot of useful and sometimes required Python and Perl scripts for certain tasks, e.g. plotting energy over the run. The setup of an AFMQC calculation requires the AFMQC-Tools from here https://github.com/QMCPACK/qmcpack/tree/develop/utils/afqmctools. However, the current version does not install them and does not handle python and perl scripts. Would be super helpful if we could add them.

However, it seems the AFMQC-Tools are more or less a real python package without a setup, not sure how to handle them best.

cc @markuskowa any suggestions on this?

sheepforce:
Adds QMCPack utility scripts in Python and Perl to the installation and ensures, that the AFQMCTools are available as Python package.

Closes #547


pkgs/python-by-name/qmcpack/package.nix
```diff
@@ -0,0 +1,98 @@
...
+  ];
+
+  # Install the python utility programs of QMCPACK
+  postInstall =
+    let
+      pythonScripts = [
+        "convert-gamess-ecp-to-qmcpack"
+        "determinants_tools.py"
+      ];
+      perlScripts = [
```
markuskowa, line 70: The perl scripts are not getting patched properly (`#!/usr/bin/env perl`). May add perl to the inputs?

pkgs/python-by-name/qmcpack/package.nix
```diff
@@ -0,0 +1,98 @@
...
+    in
+    ''
+      mkdir -p $out/${python.sitePackages}/qmcpack
+
+      for F in ${builtins.toString (pythonScripts ++ perlScripts)}; do
+        cp ../utils/$F $out/bin/.
+      done
+
+      cd ../utils/afqmctools
+      cp ./bin/* $out/bin/.
```
markuskowa, line 84:
```suggestion
      cp ../utils/afqmctools/bin/* $out/bin/.
```

Issue #379, opened 2023-12-06
Psi4 fails with missing python deps


sheepforce:
At some point during a past update the runtime of Psi4 broke. While it builds just fine, running it as an executable fails with missing numpy modules:

```bash
$ nix run github:nix-qchem/nixos-qchem#psi4
Traceback (most recent call last):
  File "/nix/store/fxmm3hgxqy29q0q0bpq1fjjf41vih223-python3.11-psi4-1.8.2/lib/psi4/driver/dependency_check.py", line 35, in <module>
    import numpy as np
ModuleNotFoundError: No module named 'numpy'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/nix/store/fxmm3hgxqy29q0q0bpq1fjjf41vih223-python3.11-psi4-1.8.2/bin/.psi4-wrapped", line 214, in <module>
    import psi4  # isort:skip
    ^^^^^^^^^^^
  File "/nix/store/fxmm3hgxqy29q0q0bpq1fjjf41vih223-python3.11-psi4-1.8.2/lib/psi4/__init__.py", line 90, in <module>
    from .driver import endorsed_plugins
  File "/nix/store/fxmm3hgxqy29q0q0bpq1fjjf41vih223-python3.11-psi4-1.8.2/lib/psi4/driver/__init__.py", line 31, in <module>
    from . import dependency_check
  File "/nix/store/fxmm3hgxqy29q0q0bpq1fjjf41vih223-python3.11-psi4-1.8.2/lib/psi4/driver/dependency_check.py", line 46, in <module>
    raise ImportError(msg)
ImportError: 
    NumPy is a runtime requirement for Psi4. Please install NumPy to proceed.

    NumPy installation with a package manager can be accomplished by the following lines:
        - conda install numpy
        - sudo yum install numpy
        - sudo apt-get install python-numpy
        - brew install numpy
```

I've no idea when exactly it happened. @markuskowa do you have any idea if something relevant in the python ecosystem in nixpkgs has changed?

markuskowa:
I do not know when this happened or why it does find numpy anymore. You could run `git bisect` to find the commit in NixOS-QChem that caused this error (and then bisect nixpkgs to nail it down further).

Maybe we should have a separate psi4 test (outside of the main derivation). That would have uncovered the error.

PR #566, opened 2024-10-14
psi4: split output -> out, dev

markuskowa:
.dev is 33M large (libint headers)!

### DIFF
--- a/pkgs/python-by-name/psi4/package.nix
+++ b/pkgs/python-by-name/psi4/package.nix
@@ -186,6 +186,8 @@ buildPythonPackage rec {
     hash = "sha256-eghnSzfbUAtYTW6wbE6KizuDujnH3Tze9zcOY7ATY60=";
   };
 
+  outputs = [ "out" "dev" ];
+
   preConfigure = ''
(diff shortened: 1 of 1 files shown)


sheepforce:
Hm, we already had this attempt and it broke Psi4; see here and here

  * https://github.com/Nix-QChem/NixOS-QChem/issues/379
  * https://github.com/Nix-QChem/NixOS-QChem/pull/380

markuskowa:
> Hm, we already had this attempt and it broke Psi4; see here and here
> 
> * [Psi4 fails with missing python deps #379](https://github.com/Nix-QChem/NixOS-QChem/issues/379)
> * [psi4: do not split output #380](https://github.com/Nix-QChem/NixOS-QChem/pull/380)

Groundhog day... However, it does not seem to break PSi4 any longer. I can run the build on a test input e.g. `result/bin/psi4 test.inp` executes just fine.

sheepforce:
Could you also try if `nix shell .#psi4` and `nix run .#psi4 -- -i test.inp` work? These were some funny problems last time. If both do work, then fine for me :)

markuskowa:
Yep, ok. That is still broken. "Great" that that the new interface behaves differently.

PR #576, opened 2024-11-06
orca: 6.0.0 -> 6.0.1

sheepforce:
Updates ORCA to its latest release with many bugfixes. Unfortunately, it introduces another bug in the build, see https://orcaforum.kofo.mpg.de/viewtopic.php?f=11&t=12483
To work around this bug, I'm patching all autoci binaries to use the correct liborca_tool library. 
Furthermore, orca now ships its own build of xtb 6.7.1. I'm overriding it with our version.

EDIT: While this works perfectly fine, the ORCA developers want to update the binaries. This will fix the issue but invalidate the hash. Maybe we better wait for the updated archives.

### DIFF
--- a/pkgs/by-name/orca/package.nix
+++ b/pkgs/by-name/orca/package.nix
@@ -18,18 +18,18 @@ assert withCustomOtoolExternal -> customOtoolExternal != null;
 
 stdenv.mkDerivation {
   pname = "orca";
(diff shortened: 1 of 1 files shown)


sheepforce:
Here we go, archives are updated and hashes are up to date, not patchelf required to fix it.

PR #601, opened 2024-12-25
staging: update flake.lock and fix fallout from gcc-14 update

markuskowa:
The latest staging-next merge, brought an update to gcc-14 and numpy-2, which broke a number of packages.

Collecting all repair work here.


sheepforce:
The ABPS/FETK fix which is required to unbreak AmberTools is here: https://github.com/NixOS/nixpkgs/pull/374252

markuskowa:
With https://github.com/NixOS/nixpkgs/pull/374380 merged and lined up, we are almost back to a clean state.

The following packages are still failing:
* pysisyphus (`FAILED tests/test_cube/test_cube.py::test_cube - AttributeError: module 'numpy' has no attribute 'product'`)
* python311.mendeleev and python312.pyastronomy. @sheepforce you fixed them for python-3.12. Are these two packages still needed for python-3.11 or can we drop from 3.11?

sheepforce:
> pysisyphus (FAILED tests/test_cube/test_cube.py::test_cube - AttributeError: module 'numpy' has no attribute 'product')

I would suggest we mark pysisyphus broken and possibly remove it. Johannes has left the group and development has thus stalled. It will now only become more and more incompatible.

> python311.mendeleev and python312.pyastronomy. @sheepforce you fixed them for python-3.12. Are these two packages still needed for python-3.11 or can we drop from 3.11?

Now that you've mentioned it, I've checked why I've packaged them at all initially. It was for `poltype2`, which I've added with absolutely insane micromamba magic instead. thus, mendeleev and pyastronomy are not actually used at all :sweat_smile: I will remove them ...

markuskowa:
Waiting now for [Hydra eval 1811136](https://hydra.nixos.org/eval/1811136) to complete (and the channel update). Should be all green then.

markuskowa:
We are all green again! I squashed some of the flake.lock commits and updated the package list.

PR #616, opened 2025-02-04
Update flake.lock/package list

markuskowa:
qchem.realease-barrier has passed


sheepforce:
Ah nice, this one contains the PRRTE fix :partying_face:

PR #641, opened 2025-03-20
turbomole: 7.8.1 -> 7.9

sheepforce:
Regular Turbomole update to 7.9 release.

### DIFF
--- a/pkgs/by-name/turbomole/package.nix
+++ b/pkgs/by-name/turbomole/package.nix
@@ -16,14 +16,21 @@
 , coreutils
 , # Configuration
   useMPI ? false
+, version ? "7.9"
 }:
+
+assert builtins.elem version [ "7.9" "7.8.1" ];
+
 let
   systemName =
     if stdenv.isLinux && stdenv.isx86_64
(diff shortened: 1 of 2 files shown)


markuskowa (changes requested):
We should make seperate versions. I just realised that we only bought "3 versions". I.e. I can not download 7.9.
Please preserve version 7.8 under turbomole78

sheepforce:
Alright, I've pushed another commit, which allows both versions.

pkgs/by-name/turbomole/package.nix
```diff
@@ -16,14 +16,21 @@
 , coreutils
 , # Configuration
   useMPI ? false
+, version ? "7.8.1"
```
markuskowa, line 19: Thanks!
Now you do not have version 7.9 in the set (i.e. `turbomole` also evaluates to 7.8.1).

sheepforce: Ah oh good point :sweat_smile: Fixed.

markuskowa (approved):

PR #657, opened 2025-04-03
multiwfn: 3.8-2024-06-14 -> 3.8-2025-03-31

### LINKED ISSUE #656: mutliwfn: replace arb with flint3
The latest `nixpkgs-unstable` yields and evaluation error: 
```
error: 'arb' has been removed as it has been merged into 'flint3'
```
This patch almost builds, but requires to update the linker flags from `-larb` to `-lflint`
```
diff --git a/pkgs/by-name/multiwfn/package.nix b/pkgs/by-name/multiwfn/package.nix
index 81eb37a..3455787 100644
--- a/pkgs/by-name/multiwfn/package.nix
+++ b/pkgs/by-name/multiwfn/package.nix
@@ -7,8 +7,7 @@
 , libGL
 , motif
 , mkl
-, arb
-, flint
+, flint3
 }:
 
 stdenv.mkDerivation rec {
@@ -35,8 +34,7 @@ stdenv.mkDerivation rec {
     libGL
     motif
     mkl
-    arb
-    flint
+    flint3
   ];
 
   enableParallelBuilding = true;
```

CC @sheepforce

sheepforce:
Fixes #656 
In nixpkgs-unstable, `flint3` is actually `flint` and `flint2` is the extra value. To be merged together with a flake update.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1743320628,
-        "narHash": "sha256-FurMxmjEEqEMld11eX2vgfAx0Rz0JhoFm8UgxbfCZa8=",
+        "lastModified": 1743689281,
+        "narHash": "sha256-y7Hg5lwWhEOgflEHRfzSH96BOt26LaYfrYWzZ+VoVdg=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "63158b9cbb6ec93d26255871c447b0f01da81619",
+        "rev": "2bfc080955153be0be56724be6fa5477b4eefabb",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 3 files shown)


markuskowa:
That fails quickly:
```
ext/2F2.c:7:10: fatal error: acb_hypgeom.h: No such file or directory
    7 | #include "acb_hypgeom.h"
```

sheepforce:
Yes, because we don't have a flake lock update yet. Should we update the lock within this PR, too?

markuskowa:
Yes, just update the flake.

sheepforce:
Should pass now.

markuskowa:
With the flake update, `qchem.pegamoid` has now a complaint:
`error: do not use python3 when building Python packages, use the generic python parameter instead`

markuskowa:
Thanks! build is clean now.

PR #675, opened 2025-05-09
flake: fix clean up package of set

markuskowa:
Closes https://github.com/Nix-QChem/NixOS-QChem/issues/674

The cleanup  leads to failure of all packages when evaluation of one fails.


sheepforce (approved):
Thank you. That was not too obvious. :smile:

PR #700, opened 2025-07-23
WIP: staging, update flake.lock

markuskowa:
* [x] Waiting on https://github.com/NixOS/nixpkgs/pull/427752 to hit `nixpkgs-unstable`
* [x] Waiting on https://github.com/NixOS/nixpkgs/pull/429313 to hit `nixpkgs-unstable`
* [x] Resolution of https://github.com/Nix-QChem/NixOS-QChem/issues/699

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1753934836,
-        "narHash": "sha256-G06FmIBj0I5bMW1Q8hAEIl5N7IHMK7+Ta4KA+BmneDA=",
+        "lastModified": 1754393734,
+        "narHash": "sha256-fbnmAwTQkuXHKBlcL5Nq1sMAzd3GFqCOQgEQw6Hy0Ak=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "8679b16e11becd487b45d568358ddf9d5640d860",
+        "rev": "a683adc19ff5228af548c6539dbc3440509bfed3",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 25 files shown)


sheepforce:
> https://github.com/Nix-QChem/NixOS-QChem/issues/699

I've played around with that already a little bit and xTB builds successfull with `enableParallelBuilding = false;` but then 3 tests fail with _severe_ numerical errors.

markuskowa:
> Hopefully meson 1.8.1 lands in nixpkgs soonish.

Meson is already at 1.8.2. 

> I've played around with that already a little bit and xTB builds successfull with enableParallelBuilding = false; but then 3 tests fail with severe numerical errors.

Since Meson is just the build system, I assume that the numerical errors are caused by something else.

markuskowa:
> I'm currently updating CP2K to 2025.2 in nixpkgs, which now needs libxc 7.0. Maybe it would be good to included that in this update cycle, as it will probably also break a few other things.

Please give it try, but as far as I remember `libxc-7.0` broke every package that depends on it. If that is still the case, we can consider keeping both 6 and 7 in parallel for the transition period. It may take a while until every package is 7.0 compatible.

sheepforce:
> We could pull in the old meson-1.7 from nixpkgs (i.e. copy the file to NixOS-QChem), and build xtb and crest with the old meson. Could this be the simplest hotfix/instant solution?

Maybe, yes. It will still require 

 * > overriding tblite to use the currently pinned commit that allows xTB to build (neither 0.4.0 nor 0.5.0)
 * > override mctc-lib 0.4.2 (breaks xTB interface) -> 0.4.1. This also requires to override jonquil -> json-fortran, again.

but it is *probably* doable. I will see if I can still get it done this week. I'm on vacation starting next week. :grimacing: :palm_tree:

PR #719, opened 2025-08-25
staging: 

markuskowa:
Fix build failures from latest nixpkgs-stagign merge.

* gamess-us: build with gfortran-14


sheepforce:
> Maybe the better solution is to replicate the Grimme packages(?).
E.g. tbliteCmake = tblite.override {};
and CP2K explicitly consumes tbliteCmake etc.
That is much uglier than the makeScope solution, but overriding would be more straightforward (maybe?).

Hm, indeed much uglier. I think for upstream nixpkgs that would be too ugly doubling all those packages without an apparent reason. But we could at least do it in our recall, maybe?

markuskowa:
I think I found a better solution:
Move the `newScope`  to the CP2K package file. That way, the messy part is self-contained. 
Downside: cp2k will have a few more inputs that are indirect since they are required for the `grimmeCmake` set.
Upside: you can feed cp2k with overridden grimme inputs as usual.
This will make it also much easier to build CP2K with CUDA support, which is now only possible if you repeat the whole `newScope` statement (?).

I will open a PR later to show this will work.

Issue #736, opened 2025-09-17
ambertools: broken on current master


markuskowa:
Ambertools is still built with python-3.11. Currently `python3.11-tkinter` is broken in nixpkgs, particularly with python-3.11 blocking a flake.lock upgrade.
Does `ambertools-25` need to be built with pyhton-3.11 or can we upgrade python here?

CC @sheepforce

sheepforce:
It does not work without modifications with Python >= 3.12, but apparently some people figured out how to get it running with 3.13, see http://archive.ambermd.org/202505/0059.html .
I'm currently very busy, but will try to draft a fix in the next days 😬

sheepforce:
Turns out to be much more complex than this. There is a fix in the easybuilders repository https://github.com/easybuilders/easybuild-easyconfigs/pull/23420 but it does more than what applies to our situation and the diff is huge. I'm not sure whether such an enormous port is worth the trouble?

sheepforce:
Alternative fix for tkinter with python 3.11 has been merged here: [#442184](https://github.com/NixOS/nixpkgs/pull/442184). When this reaches nixpkgs-unstable, AmberTools should not block the update anymore.

markuskowa:
Great! then we just wait for the next staging merge.

PR #739, opened 2025-09-24
vmd-python: 3.1.4 -> 3.1.7

sheepforce:
Updates vmd-python to the latest version, which has python 3.12 support, but unfortunately not 3.13 yet. For 3.1.7 there is no release tag for some reasons, but the tag for 3.1.4 kept moving and broke the latest version anyway.


sheepforce:
No, also AmberTools unfortunately. Unless we mention to port the fixes mentioned in #736 ...

PR #741, opened 2025-09-24
psi4: 1.9.1 -> 1.10

sheepforce:
Updates Psi4 to the latest release: https://github.com/psi4/psi4/releases/tag/v1.10

Supposedly Psi4 works now with upstream LibInt2, but honestly I have now idea how. I've tried to use upstream LibInt but got runtime errors whenever integrals were calculated. Thus, I've kept the custom libint in, as does Psi4. 

With https://github.com/NixOS/nixpkgs/pull/445777 we could enable OpenOrbitalOptimizer for Psi4 (and remove integratorxx).

### DIFF
--- a/pkgs/python-by-name/psi4/package.nix
+++ b/pkgs/python-by-name/psi4/package.nix
@@ -3,7 +3,6 @@
 , buildPythonPackage
 , buildPackages
 , isPy311
-, makeWrapper
 , fetchFromGitHub
 , fetchFromGitLab
 , fetchurl
@@ -23,7 +22,8 @@
 , gau2grid
 , libxc
 , dkh
-, dftd3
+, simple-dftd3
+, dftd4
 , pcmsolver
 , libecpint
 , cppe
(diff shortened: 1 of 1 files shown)


pkgs/python-by-name/psi4/package.nix
```diff
@@ -134,7 +135,7 @@ let
 in
 buildPythonPackage rec {
   pname = "psi4";
-  version = "1.9.1";
+  version = "1.10";
```
markuskowa, line 138: The hash needs to be update.

sheepforce: Urks, that one got lost in my git rebasing ...

sheepforce:
Alright, switched to the pregenerated libint that Psi4 is using by default and that seems to work. I've tested with the lower angular momentum version and that was working. Now building the Libint version with AM=7 ...

markuskowa:
> I can not get this running as `toPythonApplication` and I assume it also makes sense somehow. The whole idea of `toPythonApplication` is to hide the python environment at runtime and behave like a pure executable, but Psi4 tries to find and import modules in the python environment at runtime, that don't need to be present at build time. I'm thinking of removing it from top level overlay and just keep it in the python environment. What do you think, @markuskowa ?

Psi4 has an executable python script that is used to run it standalone. Thus, I would not remote from top-level. 
However, just removing the `toPyhthonApplication` is okay. If the executable does not run in that setting,  could we wrap that manually?

sheepforce:
Hm, that is interesting :smile: I'm now compiling the pre-generated libint sources in place, instead of also running the generator and then compiling. But the pregenerated libint ships with $l_\text{max} = 7$, while I've been building with $l_\text{max} = 6$ for the ERIs previously. The only alternative pregenerated version with $l_\text{max} = 5$ is unfortunately not sufficient for transition metal JK bases.
The current libint setup is the one that is also used by Psi4's releases and shipped with Anaconda. Using libcint would have been great, it's more lightweight, supports general contraction and is often faster. But here we are :grimacing: 

But even worse, I have to limit the number of build cores to ~ 4 on my 64 GiB machine. Otherwise I will run out of memory with some coupled cluster module compilation.

PR #742, opened 2025-09-24
gmda: fix hash, update commit

sheepforce:
The GDMA hash broke somehow ... I don't know how, I suspect it has something to do with `leaveDotGit = true` but as the commit didn't change I'm somewhat lost. I've updated to the most recent commit anyway.

### DIFF
--- a/pkgs/by-name/gdma/package.nix
+++ b/pkgs/by-name/gdma/package.nix
@@ -12,10 +12,10 @@ stdenv.mkDerivation rec {
   buildInputs = [ python3 ];
 
   # Needs the .git subdirectory to generate a version string.
-  src = fetchgit  {
+  src = fetchgit {
     url = "https://gitlab.com/anthonyjs/${pname}.git";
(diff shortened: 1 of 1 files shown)


sheepforce:
Hm, this seems to be a problem, too. Running this on another machine gives a different hash again ... Can you reproduce this @markuskowa ?

markuskowa:
> Hm, this seems to be a problem, too. Running this on another machine gives a different hash again ... Can you reproduce this @markuskowa ?

Since I have a huge cache, I would rarely notice things like this. It probably is caused by `leaveDotGit = true`.

sheepforce:
>  It probably is caused by leaveDotGit = true.

Yes, it probably is. But that is a problem then :sweat_smile: I would have thought that one would still get the same hash each time from different checkouts, no? Because if not, this one is broken immediately, too.

markuskowa:
Not sure; I assumed it uses submodules, which is not the case. Either it was wrong all along from a previous update or the `fetchgit`/leaveDotGit behavior changed upstream.

PR #831, opened 2026-05-08
update flake.lock, fix openblas

markuskowa:
This fix is as ugly as it gets, but it fixes the XTB build temporarily until upstream has updated `openblas-0.3.33`. This will be most likely the case in a few weeks with the next staging merge, when https://github.com/NixOS/nixpkgs/pull/513589 is merged.

However, I would not want to wait for the next staging merge to occur since this will most likely come with a new set of problems. Hence, the dirty fix for now.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1776329215,
-        "narHash": "sha256-a8BYi3mzoJ/AcJP8UldOx8emoPRLeWqALZWu4ZvjPXw=",
+        "lastModified": 1778124196,
+        "narHash": "sha256-pYEytCNic/czazbV9r3tbQ6BZzqRBg/41x2dIC5ymOo=",
         "owner": "NixOS",
         "repo": "nixpkgs",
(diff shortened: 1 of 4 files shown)


sheepforce:
Hehe indeed it's a little bit ugly. I would have voted in favour of just overriding BLAS and LAPACK with blis and libflame temporarily for xTB, which, I suppose, is fine. It's free software and redistributable and those work. I think it's somewhat easier than overriding OpenBLAS?

markuskowa:
I just checked. We only need to replace LAPACK (BLAS should be fine).
1. libflame is not a clean replacement for LAPACK. Just changing the `lapackProvider` to `libflame` does not work (we do not do that in `release.nix` in the `buildVariants` job set because it does not work well as a replacement).
2. Replacing LAPACK with `lapack-reference` does not work either in that case. 

Both options result in build failures. Namely, the problem here is that we start mixing different blas/lapack implementations in the same binary, i.e. xtb in turn depends on other libraries that consume blas/lapack. You need to replace blas/lapack for all packages in the dependency chain for this to work (i.e. practically for the whole overlay).

sheepforce:
Hm alrighty, that makes sense. Then let's go with that approach.

sheepforce (approved):

